PR #74, opened 2018-07-22
new mob values, areas fixed, meme mob placement

LarissaMayrink:


code/game/area/Space_Station_13_areas.dm
```diff
@@ -1328,6 +1328,10 @@ NOTE: there are two lists of areas in the end of this file: centcom and station
 	has_gravity = STANDARD_GRAVITY
 	requires_power = FALSE
```
AndrewL97: Can I recommend changing this value to true for /area/f13, and removing all of the requires_power = false from the sub area types?

any area that doesn't require power should then explicitly say requires_power = false, because they are the minority

PR #108, opened 2018-09-26
buffs vault door and disables weather

LetterN:


code/game/vault_door.dm
```diff
@@ -27,48 +30,70 @@
...
-//� �� ���� ������������ ��� ������  - Google translate tells me that from Russian to english this is "Do not move your arms around this way." so dont.
+/obj/structure/vaultdoor/attackby(obj/item/I, mob/living/user, params)
+	add_fingerprint(user)
+	if(istype(I, /obj/item/weldingtool) && user.a_intent == INTENT_HELP)
+		if(obj_integrity < max_integrity)
+			if(!I.tool_start_check(user, amount=0))
+				return
+
+			to_chat(user, "<span class='notice'>You begin repairing [src]...</span>")
+			if(I.use_tool(src, user, 40, volume=50))
```
ma44: I don't think there's any welder you can get normally that has a volume of 50 unless you somehow get a 80 volume industrial welding tool

code/game/vault_door.dm
```diff
@@ -27,48 +30,70 @@
...
 	set_opacity(0)
-	src.density = 0
-	is_busy = 0
+	src.density = FALSE
+	is_busy = FALSE
+	is_open = TRUE
 
 /obj/structure/vaultdoor/proc/close()
-	is_busy = 1
+	is_busy = TRUE
```
AndrewL97: Not for this pull (unless you're keen)
but we should also do a loop on everything on the door turf and gib/damage if it shuts on them, to make it a bit more risky trying to slip through as it shuts

code/game/vault_door.dm
AndrewL97: I think hacked lathes spit these out?

If this is an issue we can probably just make this a gradual thing. Like each weld uses 10 fuel and increases integrity (capped) by max_integrity/5 or something

code/modules/clothing/under/f13.dm
```diff
@@ -188,7 +195,7 @@
 
 /obj/item/clothing/under/f13/cowboyg
 	name = "Cowboy outfit"
-	desc = "A dirt Grey cowboy outfit. Specially usefull for herding brahmins."
+	desc = "A dirt grey cowboy outfit. Specially usefull for herding brahmins."
```
AndrewL97: gey*

code/game/vault_door.dm
AndrewL97: @ma44 and @LetterN 

I am fine with this change. Up to you lads if incremental repairs or something else should be used instead.

PR #114, opened 2018-10-31
vault updates

LetterN:
fixes mapping issues
makes turf less broken


AndrewL97:
@LetterN stop merging straight into Master without time for a review please. If its been a week, go for it, but otherwise at least one person should go over it.

PR #115, opened 2018-11-02
Adds 3 minute respawn timer + bypass for admins, refactors factions for jobs so it isn't horrible

ma44:


code/modules/mob/mob.dm
```diff
@@ -399,28 +399,80 @@
 		to_chat(usr, "<span class='boldnotice'>You must be dead to use this!</span>")
 		return
 
-	log_game("[key_name(usr)] used abandon mob.")
+	var/is_admin = check_rights_for(src.client, R_ADMIN)
```
AndrewL97: Hey @ma44 , are you able to refactor this to cut down some of the duplicate code? There are a few death checks and other bits that are repeated a few times.

Something like this, its untested but you should get the gist of it. Ping me if questions

```
/mob/verb/abandon_mob()
	set name = "Respawn"
	set category = "OOC"
	
	/* check respawn is on */
	if (CONFIG_GET(flag/norespawn))
		return
	/* check player is actually dead */
	if ((stat != DEAD || !( SSticker )))
		to_chat(usr, "<span class='boldnotice'>You must be dead to use this!</span>")
		return
	
	/* if player has no body, allow instant respawn, otherwise do standard checks */
	if(src.mind.current)
		var/is_admin = check_rights_for(src.client, R_ADMIN)
		var/deathtime = world.time - mind.current.timeofdeath //How long dead for in deciseconds
		/* check if the respawn cooldown has expired, and check for admin override if not */
		if(deathtime < 1800)
			to_chat(src, "You've been dead for [deathtime / 10] seconds. You must be dead for at least three minutes to respawn.")
			if(is_admin) /* if player is an admin, and cancels the override, return */
				if(alert("Normal players must wait at least 3 minutes to respawn! Continue?","Warning", "Respawn", "Cancel") == "Cancel")
					return
				else /* admin chose to override, so log it rather than returning */
					log_game("[key_name(usr)] used abandon mob while bypassing the regular death cooldown VIA admin prompt.")
			else /* if player is not an admin, they can't override, so return */
				return
	/*end src.mind.current - we survived the various checks, so perform the actual respawn */
	log_game("[key_name(usr)] used abandon mob.")

	to_chat(usr, "<span class='boldnotice'>Please roleplay correctly!</span>")

	if(!client)
		log_game("[key_name(usr)] AM failed due to disconnect.")
		return
	client.screen.Cut()
	client.screen += client.void
	if(!client)
		log_game("[key_name(usr)] AM failed due to disconnect.")
		return

	var/mob/dead/new_player/M = new /mob/dead/new_player()
	if(!client)
		log_game("[key_name(usr)] AM failed due to disconnect.")
		qdel(M)
		return

	M.key = key
	return	
```

AndrewL97:
Added fix for ghost/gib/brain not picking up respawn timer properly.
tested with excessive drop-bomb, turf changes are good

will merge as soon as travis is done

thanks matey

MrBobCool:
No stop, no timer. GOD NO.

AndrewL97:
timer is being totally abused for shittery, don't be a dingus

PR #142, opened 2018-11-05
fixes more access issues, also made sure vaulties cant buy a different shuttle

LetterN:
_stop buying scrapheaps ree_


LetterN:
>check fails because mappdiffbot cant upload the map diffs

PR #153, opened 2018-11-08
Techweb fix available node issue

AndrewL97:
## Description

issue found with rnd - once vault researches a node, no other techweb can research it

## Motivation and Context

makes this weird one sided race. If vault rushes the early nodes, it kills bos.

## How Has This Been Tested?

researched node in BOS
confirmed node can be researched in vault
researched another node in vault
confirmed node can be researched in BOS

n/a


AndrewL97:
github is displaying a weird diff for this - will close and push a cleaner branch

PR #155, opened 2018-11-08
Allows changing of protolathe's tech webs host research with tech disks

ma44:
Allows changing of techwebs that departmental protolathes uses to sync and get designs from by slapping the protolathe with a tech disk. 
## Description

This fixes the issue of default departmental protolathes not being limited to just the global 'unknown' techweb, which was mainly a problem for vaulters not being able to use the science they researched at the protolathes like med, security and service.
## Motivation and Context

So the damm code can work

## How Has This Been Tested?

It hasn't; hope it works

### DIFF
--- a/code/modules/research/machinery/_production.dm
+++ b/code/modules/research/machinery/_production.dm
@@ -331,3 +331,10 @@
 
 	l += "</tr></table></div>"
 	return l
+
+/obj/machinery/rnd/production/attackby(obj/item/I, mob/user)
+	if(istype(I, /obj/item/disk/tech_disk))
+		var/obj/item/disk/tech_disk/disk = I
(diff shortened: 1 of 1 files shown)


AndrewL97:
Can you test this works and let me know?
Test notes like how to get the techweb linked to the disk, etc. would be cash

ma44:
To get the techweb linked to the shit, you 'download' the data from a RD console onto the disk VIA inserting the disk into the console

I'll have to make something like the RD console being able to get say a techweb from the starting nearby unique server, as ATM if you upload a newly built techweb onto a unique techweb RD console, it would make the RD console into the global tech web and forever erasing the BoS/Vault tec techweb from existence unless an admin spawns in a subtype of a techdisk

PR #159, opened 2018-11-09
Checks for null mind when respawning

AndrewL97:
## Description

allows observers to respawn 3 minutes after round start - fixes runtimes

## Motivation and Context

so observers can respawn

## How Has This Been Tested?

Observed, waited for respawn, was able to respawn
Spawned as normal job, used suicide to die, was able to respawn afterwards
spawned as normal job, gibbed, was able to respawn
spawned as normal job, ghosted, was able to respawn

### DIFF
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -402,7 +402,7 @@
 		return
 
 	/* if player has no body, allow instant respawn, otherwise do standard checks */
-	if(src.mind.current)
+	if(src.mind && src.mind.current) /* check if mind is null, as observers don't have one - they can instant respawn */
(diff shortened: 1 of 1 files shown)


code/modules/mob/mob.dm
```diff
@@ -402,7 +402,7 @@
 		return
 
 	/* if player has no body, allow instant respawn, otherwise do standard checks */
-	if(src.mind.current)
+	if(src.mind && src.mind.current) /* check if mind is null, as observers don't have one - they can instant respawn */
```
rhicora: This looks great, it won't fall over if they don't have a mind.

PR #162, opened 2018-11-09
Adds locks

UndeadNappist:
Re-adds barricading doors.

Adds a craftable lock item that can be applied to the wooden doors. Each new lock has a different UID. When applied to a door, a key spawns that can unlock anything with the same UID. Set new locks to be the same UID by applying the lock to a door which already has a lock, but is unlocked. Apply to a new door to spawn another key.

This is a pile of if statements and could probably be handled better, but it works.

image

image


ma44:
Adding a way to get new locks (crafting?) and could also probably just be limited to wooden doors and not every single possible subtype of door.

UndeadNappist:
You can already get new locks. It is a hand-crafting steel sheet recipe.

I don't see the issue with having it be every type of door, since they all can be smashed down. They're less effective than airlocks for actually keeping people out.

There is a concern with the fact locks get deleted on use, so potentially someone could use their locks, someone could make a new one, apply it, and then the key would retroactively work on all the previous locks.

Not quite sure how to go about stopping that.

code/game/objects/structures/mineral_doors.dm
```diff
@@ -71,6 +73,9 @@
 		return
 	if(isliving(user))
 		var/mob/living/M = user
+		if((/obj/structure/barricade in src.loc)||(lock == 2))
+			M << "It won't budge!"
```
LetterN: to_chat(M, "<span class='warning'>It won't budge!</span>"

code/game/objects/structures/mineral_doors.dm
```diff
@@ -128,6 +136,35 @@
 		icon_state = initial_state
 
 /obj/structure/mineral_door/attackby(obj/item/I, mob/user, params)
+	if(istype(I, /obj/item/lock) && do_after(user, 5, target = src))
+		if(lock == 1)
+			user << "<span class='notice'>You key the lock to be the same.</span>"
```
LetterN: to_chat(user, "<span class='notice'>You key the lock to be the same.</span>")

code/game/objects/structures/mineral_doors.dm
```diff
@@ -128,6 +136,35 @@
...
+			uid = src.uid
+		return
+		if(lock == 2)
+			user << "<span class='notice'>This door already has a lock on it!</span>"
```
LetterN: to_chat this too

code/game/objects/structures/mineral_doors.dm
```diff
@@ -128,6 +136,35 @@
...
+	src.uid = uid
+	lock = 1
+	spawn(/obj/item/key)
+		uid = src.uid
+	qdel(/obj/item/lock)
+	user.visible_message("[user] adds a lock to the door.")
+	return
+	if(istype(I, /obj/item/key))
+		if(lock == 0)
+			user << "<span class='notice'>This door doesn't have a lock.</span>"
```
LetterN: to_chat this too

code/game/objects/structures/mineral_doors.dm
```diff
@@ -128,6 +136,35 @@
...
+			return
+		if((src.lock > 0) && (uid != src.uid))
+			user << "<span class='notice'>This is the wrong key!</span>"
```
LetterN: to_chat

UndeadNappist:
Extensive testing has revealed a of interaction issues. Closing until it's fixed.

code/game/objects/structures/mineral_doors.dm
```diff
@@ -81,6 +86,9 @@
 			else
 				SwitchState()
 	else if(ismecha(user))
+		if(lock == 2)
```
AndrewL97: Can we get the lock states (0,1, and 2) a predefine?
Like LOCK_BROKEN LOCK_UNLOCKED and LOCK_LOCKED? Just to make maintenance easier

Other than that (and the other requests), totally solid idea, thanks for the work

UndeadNappist:
Here we go again.

PR #163, opened 2018-11-09
A couple fixes

rhicora:
- "Capture" camp followers now simply have the "nothing" job and no special gear
- Non-voters will not be counted for either side in restart votes
- Reorganized Legion kitchen, added a working microwave and meatspike
- Gives Legion Scouts and Explorers the correct form of ammo in their bag

image


rhicora:
Solves https://github.com/BadDeathclaw/TG-Claw/issues/164

PR #176, opened 2018-11-11
Add buildable crosses to throw degenerates on

rhicora:
- Adds handcrafting item "cross", which builds a cross for 10 wood
- Deconstructing crosses with a crowbar yields the wood back
- Add interactions to buckle and unbuckle mobs
- Add the cross sprite and sound effect

image

PR #177, opened 2018-11-11
The Casino Update

rhicora:
This adds a casino to the Den, and reorganizes the Den quite a bit.

It also adds back caps + legion money and makes them pressable. NCR dollars are in!

image

Dependency: this map was forked from `degenerates-like-you` (#176)


code/game/objects/structures/kitchen_spike.dm
```diff
@@ -149,4 +149,84 @@
...
+	icon = 'icons/obj/cross.dmi'
+	icon_state = "cross"
+	desc = "Degenerates like you belong on one of these."
+	density = 1
+	anchored = 1
+	buckle_lying = 0
+	can_buckle = 1
+	bound_height = 64
+
+/obj/structure/kitchenspike/cross/attackby(obj/item/I, mob/user, params)
```
AndrewL97: Put this check as a crowbar_act instead of checking crowbar in attackby. Look at the current one in kitchen_spike.dm

code/modules/mining/ores_coins.dm
```diff
@@ -378,13 +382,43 @@ GLOBAL_LIST_INIT(sand_recipes, list(\
...
 /obj/item/coin/iron
-	name = "iron coin"
+	name = "bottle cap"
+	desc = "A humble bottle cap. This one seems to be of the Nuka Cola variety."
 	cmineral = "iron"
 	icon_state = "coin_iron_heads"
-	value = 1
+	is_non_flippable = TRUE
+	sideslist = list("heads")
+	value = 2.5
```
AndrewL97: for coin values, please make defines for them like COIN_VALUE_IRON, COIN_VALUE_CAP, etc. pref using cap as 1, with everything else being defined based off that
like
```
#define COIN_VALUE_CAP 1
#define COIN_VALUE_NCR 0.25 * COIN_VALUE_CAP
```

code/modules/mining/ores_coins.dm
```diff
@@ -324,6 +324,8 @@ GLOBAL_LIST_INIT(sand_recipes, list(\
 	var/cooldown = 0
 	var/value = 1
 	var/coinflip
+	var/is_stack = 0
```
AndrewL97: You're going to hate me for this, but

rather than making a second stack type class, with its own unique handling, just make coins a subclass of stacks
you will need to update things (like vendors) that currently take coins, and have them take 1 from the stack amount

then in that subclass, add the flipping, etc if there is only 1 coin

painful now, but it saves us from having two stack classes to maintain

code/game/objects/structures/kitchen_spike.dm
rhicora: Ok.

code/modules/mining/ores_coins.dm
rhicora: Ok.

code/modules/mining/ores_coins.dm
rhicora: Heck.

rhicora: I kinda didn't want to use base stacks. They only have 3 variants (I want to use 6) and have a bunch of material construction related baggage, that's why I avoided them. I would also have to change the coin press machine to make stacks for some coin types and single coins for others. Idk about this...

rhicora: image

rhicora: image

AndrewL97: Ill have a squizz when home and see what we can do

worst case is we can make a stack/currency/ child class, with a return override for the various crafting things.

just dont want to double up on the various duplication, merging, etc. handling

rhicora: Makes sense. Good luck!

code/game/objects/items/stacks/cash.dm
```diff
@@ -13,59 +14,75 @@
...
+	//Used to generate a random stack of dosh
+	var/min_dollars = 1
+	var/max_dollars = 500
 
+/obj/item/stack/spacecash/random/New()
+	var/obj/item/stack/spacecash/randy = new //makes new stack
+	randy.loc = src.loc //sets stack location to randstack location
+	randy.amount = rand(min_dollars, max_dollars)
+	randy.update_icon()
+	del src
```
LetterN: qdel(src)

code/game/objects/structures/kitchen_spike.dm
```diff
@@ -149,4 +149,93 @@
...
+			L.loc = src.loc
+			L.emote("scream")
+			L.adjustBruteLoss(10)
+			L.buckled = src
+			L.dir = 2
+			buckle_mob(L, force=1)
+			L.pixel_y = 26
+			L.overlays += image('icons/obj/cross.dmi', "lashing")
+			return
+		user << "<span class='danger'>You can't use that on the cross!</span>"
```
LetterN: to_chat()

code/game/objects/structures/kitchen_spike.dm
```diff
@@ -149,4 +149,93 @@
...
+
+		else
+			M.visible_message(\
+			"<span class='warning'>[M.name] struggles to break free from the [src]!</span>",\
+			"<span class='notice'>You struggle to break free from the [src], exacerbating your wounds! (Stay still for two minutes.)</span>",\
+			"<span class='italics'>You hear violent scraping and struggling.</span>")
+			M.adjustBruteLoss(20)
+			if(!do_after(M, 1200, target = src))
+				if(M && M.buckled)
+					M << "<span class='warning'>You fail to free yourself!</span>"
```
LetterN: to_chat()

AndrewL97:
@rhicora 

TODO:
Remove/revert old coins back to silver, gold, iron, etc.
fix conflicts

Test the slot machines a bit
 - I was able to test adding coins and new money, and spinning before BYOND died on me, just needs some more poking

Test the coin press a little more
It was working, but second pair of eyes is good
May want to also remove the moneybag from spawning when making coins, because otherwise you can easily magic up large amounts of bags

rhicora:
@LetterN Done.

rhicora:
@AndrewL97 Tested! I'm not concerned with people making lots of bags, eventually I will have them craftable out of leather.

rhicora:
Aaaallll right, ready to go @AndrewL97

Issue #179, opened 2018-11-12
Two Decanus Roles


companoo:
Basically, a Centurion should have a clear 2IC, Decanus sort of fills that role but the fact there are three of them undermines it. I propose that similar to Legionary and Veteran Legionary in-game, a Veteran Decanus role is created, with it as the deputy to the Centurion, they can be distinguished by veteran armor? + maybe a new vet Decanus headpiece, which could be sprited with relative ease by recolouring some feathers in the current sprite. This would enhance the legion chain ofcommand, making it relatively like the NCR's, ie, 2sergeants-lieutenant-captain, and on legion side 2recruitdecanus>1vetdecanus>cent

RenGusta:
Just to make this abundantly clear, Vexillarius isn't a 2IC, they're a Flag Bearer and source of Morale.

Operativ:
I agree with you there, in both Legion's lore fluff and in the Roman Empire they were flag-bearers. I just meant that the game points out that the Vexillarius is a "special guy" in the Legion (not special in a bad way! :P ).

A realistic equivalent of a Lieutenant would be something that the Roman legion knew as an "Optio". They could use a true 2IC there.

Operativ:
It's true that they don't utilise the full Roman legion ranks, unlike NCR who has a very wide span of the army ranks.

Either a Veteran Decanus (which indeed do appear in the universe, wearing a higher quality, veteran armour as mentioned by @companoo here), or perhaps a Praetorian (soldiers and officers with the potential to become a Centurion in the future) - although the Praetorians are more of a fistfighter in the fluff, and as we know fighting with fists in SS13 isn't the brightest idea without some kind of glove weapon.

If it falls on a Veteran Decanus, then it's going to be an easy addition, however a Praetorian might need a unique glove weapon that boosts unarmed damage.

If we don't count Vexillarius for aiding the Centurion (because they aren't really leading material, more of inspirational flag-bearing people), then my vote would be on a Veteran Decanus as it's rather simple to implement.

roboticxlink:
This was veteran decanus yes? This can be closed?

PR #186, opened 2018-11-13
Veteran Decanus & Laser/Plasma weaponry rebalance

Operativ:
- Added 1 Veteran Decanus for the Legion,
- Decreased the number of Decanus from 3 to 2 to compensate for the above change,
- Both roles have different helmets, the old Decanus helmet became Veteran Decanus helmet (which is a slightly more resistant version of the base helmet), and the Decanus has received a resprited helmet that does not have the white feathers,
- Added a Veteran Decanus spawn point on the map, near the Centurion's office,
- Changes to the Laser and Plasma weaponry damage and energy consumption.

This addresses the #179 by adding a second-in-command of the Legion faction to give a similar depth of organisation as with the NCR (two sergeants and a lieutenant, where the Legion had three decanus before).


Operativ:
@AndrewL97 
Changes added as requested! It seems I've unintentionally omitted these three files. Thank you for the heads-up!

AndrewL97:
Tested, job spawns work, job bans updated, etc.

PR #190, opened 2018-11-14
More hair options

rhicora:
Adds 5 new hair options.

image


.gitignore
```diff
@@ -5,6 +5,7 @@
 
 #Ignore byond config folder.
 /cfg/**/*
+/config/admins.txt
```
AndrewL97: ahhhhh

AndrewL97:
damn that was fast

Are all the new hair options available in character creation? @rhicora

rhicora:
You bet.

.gitignore
rhicora: This file shouldn't be committed, it is individual for whoever is hosting a test server. The one in the repo works for you but doesn't work for everyone else.

PR #191, opened 2018-11-14
Adding limited loadout for Pusher role

roboticxlink:
## Description
Adding a Duster, 3 Jet, and 2 Med-x to the Pusher load out instead of it just being a box.

Also fixed a typo found for Mentats

## Motivation and Context
Pusher had box only for spawn

## How Has This Been Tested?
spawned as pusher. Confirmed loadout worked.


AndrewL97:
solid

PR #200, opened 2018-11-16
Sleeper chems fix

roboticxlink:
## Description
-Higher tier chems weren't injecting
-adds fix to add chems to button data list

## How Has This Been Tested?
tested injecting chems


code/game/machinery/Sleeper.dm
```diff
@@ -44,6 +44,7 @@
 	available_chems = list()
 	for(var/i in 1 to I)
 		available_chems |= possible_chems[i]
+		chem_buttons |= possible_chems[i]
```
AndrewL97: Hey @roboticxlink 
this is spewing runtimes (check the travis details), because chem_buttons hasn't been initialized yet when doing the default parts proc (its done in reset_chem_buttons)

reset_chem_buttons should also be doing what you're trying to achieve

Can I suggest doing some to_chat debugging to find out what reset_chem_buttons isn't doing what it should be doing?

roboticxlink: Yeah, I can take a look.

Not sure what the travis details are though. So I'll check what's up in reset_chem_buttons

roboticxlink: Er, so, I just took a look with a to_chat, and it said it added them all to the list in reset_chem_buttons, and was working fine. And it doesn't look like my other change is in the one I'm using soooo... Uh, is it fixed now then? 

I'll try handupgrading it instead of just tweaking component ratings, see if that's an issue.

roboticxlink: Yeah, issue comes from handupgrading. Looks like it reset_chem_buttons to the first 5 twice, but the buttons still show up. Hmmmm. Is the component rating supposed to be 16? 
It looks like they go up by powers of 2 (1,2,4,8,16), and the code looks like it just looks at their rating?

But that should still be over and above, not missing them...

AndrewL97:
Hey @roboticxlink 
found the issue
there was a recent change that modified the efficiencies to those stock parts like you said, they are now 1,2,4,8,16

This shouldn't have happened, but I missed the change. Can you please change these back to 1,2,3,4?
https://github.com/BadDeathclaw/TG-Claw/blob/684a9fff50f8fc30a0552e74388cd10db499c746/code/modules/research/stock_parts.dm

image

these ratings should reflect the actual rating level, i.e rating 2 should be 2, not 4, etc.

let me know if you have issues.

This should be the only fix needed (hopefully), and should also fix the new bugs like #210

roboticxlink:
Yes! I knew that! I just didn't know if that would cause other issues or what. I'll go through and fix those when I get home from work.

roboticxlink:
Closing this, should be fixed with PR #215

PR #201, opened 2018-11-16
Update living.dm

UndeadNappist:
-Creates a soft protection against mobs in spawn locations by spawning players asleep, because they ignore unconscious players.
-Nerfs heavy sleeper trait indirectly

## How Has This Been Tested?
Late join and new join players both spawn with mob_sleep() triggered.


AndrewL97:
I don't like this change,
its weird that everyone sleeps for ~30 seconds or so on round start
its just more annoying, and mobs will attack you when you wake up anyway

UndeadNappist:
Mobs take a bit of time to re-target you to wake up. This was just supposed to cover people who were lagging or loading in.

PR #204, opened 2018-11-17
Job exp system

LetterN:
does work, DB screenshot bellow 

image


AndrewL97:
Does this require any config on the server? If not, happy to smash this in and update today

LetterN:
the config changes are already in there, also admins can bypass this, and admins can select people to bypass it also.

AndrewL97:
siq

AndrewL97:
Bonus points for screenshot

PR #207, opened 2018-11-18
3dog fixes

rhicora:


code/modules/client/client_procs.dm
```diff
@@ -1,7 +1,7 @@
 	////////////
 	//SECURITY//
 	////////////
-#define UPLOAD_LIMIT		1048576	//Restricts client uploads to the server to 1MB //Could probably do with being lower.
+#define UPLOAD_LIMIT		5242880	//Restricts client uploads to the server to 5MB //Could probably do with being lower.
```
LetterN: oh god, time to play 5mb of midis

rhicora: A typical ogg will be something like 1.2 MB so this avoids it being cut off by a limit of one.

AndrewL97: For the sake of server sanity, can we revert this back to 1MB? 
All sorts of fun things like somebody trying do download resources on connect, people with slow internet, etc. can be donked by this. Us poor Australians are lucky to get 1mbs

rhicora: This feature basically doesn't work without 2MB at least. I can set it there if need be.

AndrewL97: bugger it, if it becomes an issue we will just revert it to 2MB, then 1MB again

PR #217, opened 2018-11-21
Invisible icon fix

UndeadNappist:
Some cleanup and fixes, adding icon states, adding the ability to craft soap in the wild, etc.


code/modules/clothing/head/collectable.dm
```diff
@@ -4,6 +4,7 @@
 /obj/item/clothing/head/collectable
 	name = "collectable hat"
 	desc = "A rare collectable hat."
+	icon_state = "viking"
```
LetterN: this should fallback into the default "no-texture" texture

lewd/objects.dm
```diff
@@ -73,7 +73,8 @@
 	//var/blood_DNA = list()
 	beauty = -50
 
-/obj/effect/decal/cleanable/cum/New()
+/obj/effect/decal/cleanable/cum/Initialize()
 	..()
 	dir = pick(1,2,4,8)
+	reagents.add_reagent("cum", rand(8.13))
```
LetterN: >'.' instead of ,

LetterN: rand(8,13)

code/modules/clothing/head/collectable.dm
UndeadNappist: It spawns on raiders sometimes as an invisible hat, and it's annoying. Either this or removing it from the spawn chance.

LetterN:
Just fix the conflicts and it's good to go!

PR #218, opened 2018-11-21
Explosions will no longer make heat

ma44:
## Description

Explosions will never make turfs heated
## Motivation and Context

Stops atmos lag caused by hotspots, in addition because of all of the lag, hotspots would also make the area constantly do fire damage usually forever
## How Has This Been Tested?

By blowing up things and seeing it doesn't create heat and no excessive atmos lag caused by it

### DIFF
--- a/code/datums/explosion.dm
+++ b/code/datums/explosion.dm
@@ -180,7 +180,7 @@ GLOBAL_LIST_EMPTY(explosions)
 				Trajectory = get_step_towards(Trajectory, epicenter)
 				dist += cached_exp_block[Trajectory]
 
-		var/flame_dist = dist < flame_range
+		//var/flame_dist = dist < flame_range
 		var/throw_dist = dist
 
 		if(dist < devastation_range)
--- a/code/datums/explosion.dm
+++ b/code/datums/explosion.dm
@@ -203,9 +203,9 @@ GLOBAL_LIST_EMPTY(explosions)
 				var/atom/A = O
 				if(!QDELETED(A))
 					A.ex_act(dist)
-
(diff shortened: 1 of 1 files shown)


roboticxlink:
Will this fix thec ground fires not burning out?

UndeadNappist:
Bonfires generate hotspots as well, as do dragonfire shotguns.

roboticxlink:
Bonfires make sense. I haven't seen issues with them too much unless you stand right next to one. But I've had some good rp someone blindfolding me and bringing me next to the fire and could feel the heat in my lungs. I vote to keep hot bonfires if possible.

ma44:
@roboticxlink it won't fix ground fires being burnt out, the desert tiles and outside should filter out the air like lavaland, but if atmos gets big enough then it will take a while to fix things. Sparks also generate heat

UndeadNappist:
Forever or instantly. Burning squares from Dragonfire shells appear to stay burning, but have this weird effect where adjacent squares or squares that had hotspots have messed up kpa and are actually cold.

AndrewL97:
Ill try to do some tests, I think the bug should be fixed in atmos/planetary_atmos, this just stops one of the symptoms

UndeadNappist:
Recent tests with dragonfire shells reveal infinitely burning fires no long occur outside, so this was fixed somewhere along the line. This is now a nonissue except indoors (as it should be).

rhicora:
Since this was last described as a nonissue / something to be fixed in planetary_atmos, I'm going to close it for the moment. Feel free to reopen this if you think it'd be valuable.

PR #221, opened 2018-11-21
[WORK IN PROGRESS] Adds in a framework for hardsuits to become power cell eating power armor

ma44:
## Description

Makes a new kind of hardsuit that takes powercells as well as consume them for power
## Motivation and Context

Because rigsuits is absolute horrible to deal with, may as well make a new one with no bad vars or whatever
## How Has This Been Tested?

Still not fully tested


AndrewL97:
approved for a week, going to mergy merge

PR #228, opened 2018-11-23
Carp meat toxicity increase

imw-Oculus:
## Description
This change increases the amount of carpotoxin in Cazador meat by three.

## Motivation and Context
My motivation in creating this push is to make more reagents available for Rezadone and Zombie Powder creation. It's also, frankly, to test out pushes and ensure that everything is functioning as it should.

## How Has This Been Tested?
I did manage to get my private server up and running - and the change is included, as per grinding Cazador meat.

### DIFF
--- a/code/modules/food_and_drinks/food/snacks_meat.dm
+++ b/code/modules/food_and_drinks/food/snacks_meat.dm
@@ -16,13 +16,13 @@
 	foodtype = MEAT
 
 /obj/item/reagent_containers/food/snacks/carpmeat
-	name = "carp fillet"
-	desc = "A fillet of spess carp meat."
+	name = "cazador fillet"
+	desc = "A fillet of cazador meat."
 	icon_state = "fishfillet"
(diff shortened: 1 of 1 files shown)


imw-Oculus:
Ah, there we go.

imw-Oculus:
That was my thought. Carpotoxin and Rezadone are pretty expensive to produce, so bumping up the amount of Carpotoxin in the meat seems to be a reasonable balance.

code/modules/food_and_drinks/food/snacks_meat.dm
```diff
@@ -19,7 +19,7 @@
 	name = "carp fillet"
 	desc = "A fillet of spess carp meat."
```
AndrewL97: Rename so its cazadore related, and ill approve

imw-Oculus: Done and done

PR #229, opened 2018-11-23
Lifegiver trait

imw-Oculus:
## Description
Adds the Lifegiver perk, as per Fallout. Increases maximum health by 5 (15, per suggestion). Cost for this perk is 3. My only concern is balance.

## Motivation and Context
A request made on server for the lifegiver perk. 

## How Has This Been Tested?
maxHealth variable change confirmed in private testing.

### DIFF
--- a/code/datums/traits/good.dm
+++ b/code/datums/traits/good.dm
@@ -9,8 +9,6 @@
 	gain_text = "<span class='notice'>You feel like you could drink a whole keg!</span>"
 	lose_text = "<span class='danger'>You don't feel as resistant to alcohol anymore. Somehow.</span>"
 
-
-
 /datum/quirk/apathetic
 	name = "Apathetic"
 	desc = "You just don't care as much as other people. That's nice to have in a place like this, I guess."
@@ -27,8 +25,6 @@
 	if(mood)
 		mood.mood_modifier = 1 //Change this once/if species get their own mood modifiers.
 
-
-
 /datum/quirk/drunkhealing
 	name = "Drunken Resilience"
 	desc = "Nothing like a good drink to make you feel on top of the world. Whenever you're drunk, you slowly recover from injuries."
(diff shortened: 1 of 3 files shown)


imw-Oculus:
Thoughts on whether it should be an increase of 15 health? More? Less or more cost for the perk?

roboticxlink:
> Thoughts on whether it should be an increase of 15 health? More? Make the quirk cost less/more?

More for the cost. But I would add a few other perks, too, otherwise it'll just be 'meta' for everyone to take this one, especially with the freebie heirloom trait.

imw-Oculus:
I'd be happy to add more perks, but I thought I'd  do it one at a time for atomicness' sake.

roboticxlink:
I'd say do a handful and add them to the same branch/etc. Some good negs and pros, and not just necessarily one at a time. Just because such a new feature, at the moment, I don't think perks do TOO much... Like poor aim trait, etc.

UndeadNappist:
Try to limit yourself not to traits that directly impact stats like health or speed unless under strict circumstances, and those that influence mood or indirectly change how you approach and tackle the game.

E.g., using game mechanics to encourage roleplay and different playstyles.

UndeadNappist:
Fatman trait: "When you've eaten yourself to satiation, you have damage resistance," which would be a similar effect with drawbacks or some interesting back-and-forth. Since you'd likely become fat and slow in the process.

Stuff like that.

imw-Oculus:
Sounds good to me, I'll keep that in mind for the future. And fatman doesn't sound like a bad perk, I'll code that up tomorrow.

code/datums/traits/good.dm
```diff
@@ -47,16 +42,25 @@
...
+	name = "Lifegiver"
+	desc = "You embody wellness! Instantly gain +5 maximum Health"
+	value = 3
+	mob_trait = TRAIT_LIFEGIVER
+	gain_text = "<span class='notice'>You feel more healthy than usual.</span>"
+	lose_text = "<span class='danger'>You feel less healthy than usual.</span>"
 
+/datum/quirk/lifegiver/on_spawn()
+	var/mob/living/carbon/human/mob_tar = quirk_holder
+	mob_tar.maxHealth += 15
```
LetterN: "+5 maximum Health"
maxHealth += 15

LetterN: make the description more accurate

imw-Oculus: Oof, didn't see that. Fixing now.

imw-Oculus:
That should do it!

PR #230, opened 2018-11-24
Nerfs Garborgs

rhicora:
- Removes the VTEC upgrade, which made borgs insano drano fast.
- Removes the size upgrade, which made borgs ugly as sin.
- You mop dry.

### DIFF
--- a/code/game/objects/items/mop.dm
+++ b/code/game/objects/items/mop.dm
@@ -39,7 +39,7 @@
 		return
 
 	if(reagents.total_volume < 1)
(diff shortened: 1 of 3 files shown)


ma44:
Approve; previous iterations of fallout had vtec/borg movement speed heavily reduced, borg expand is a good meme tho keep it for admin spawn only :ok_hand:

AndrewL97:
There is a transform command in the view-variables dropdown
lets you rotate/resize players

just use that if you need to commit buggery

PR #236, opened 2018-11-25
Updates healing powder

UndeadNappist:
-Healing_powder now calls on hallucination proc, not druggy
-Has an OD effect

### DIFF
--- a/code/modules/reagents/chemistry/reagents/medicine_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/medicine_reagents.dm
@@ -1282,15 +1282,22 @@
 	color = "#A9FBFB"
 	taste_description = "bitterness"
 	metabolization_rate = 0.25 * REAGENTS_METABOLISM
+	overdose_threshold = 30
 
 /datum/reagent/medicine/healing_powder/on_mob_life(mob/living/M)
 	M.adjustFireLoss(-3*REM)
 	M.adjustBruteLoss(-3*REM)
-	M.druggy = max(M.druggy, 5)
+	M.hallucination = max(M.hallucination, 5)
 	if(prob(7))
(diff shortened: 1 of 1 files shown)


imw-Oculus:
This will pretty severely gimp Legion. Hallucinations are brutal.

code/modules/reagents/chemistry/reagents/medicine_reagents.dm
```diff
@@ -1282,16 +1282,23 @@
 	color = "#A9FBFB"
 	taste_description = "bitterness"
 	metabolization_rate = 0.25 * REAGENTS_METABOLISM
+	overdose_threshold = 30
 
 /datum/reagent/medicine/healing_powder/on_mob_life(mob/living/M)
 	M.adjustFireLoss(-3*REM)
 	M.adjustBruteLoss(-3*REM)
-	M.druggy = max(M.druggy, 5)
+	M.hallucination = max(M.hallucination, 5)
```
AndrewL97: My only request is to put this behind a prob (like the giggle)
otherwise it might get old fast, and too predictable

AndrewL97: maybe even chuck it in the giggle prob() section, to save on probs

UndeadNappist: Having tested the hallucination, and the huge number of effects it can have - including simply noises, and the random nature of them - I can outright state that's incorrect.

I am more inclined to re-add druggy and force people to suffer through both effects due to how inconsistent hallucination is already.

AndrewL97: Looked at handle_hallucinations(), and the way this will work should be perfect
Ill take off my grumpy old man hat

PR #252, opened 2018-12-01
Bodies disintegrate under excessive damage

UndeadNappist:
After taking an excessive amount of burn, brute, or toxin damage, your body will be destroyed in a variety of methods.

Bodyparts can now take a hideous amount of damage as part of the update instead of arbitrarily capping.

Bodyparts causing their owner to die because you inflicted 1000 points of damage to the foot is Not A Thing Anymore, due to limbs being sawn off when you hit them enough.

[Not Recommended]: It seems like you might actually be able to increase a person's current health by taking off injured limbs, thus buying them more time in critical condition.

### DIFF
--- a/code/modules/mob/death.dm
+++ b/code/modules/mob/death.dm
@@ -9,5 +9,10 @@
 /mob/proc/dust()
 	return
 
+//Proc for bodies liquifying.
+//I don't know why everything is so split up like this, it's horrible, but the bulk of these three procs is buried in mob/living/death.
+/mob/proc/liquefy()
+	return
+
 /mob/proc/death(gibbed)
 	return
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -512,8 +512,15 @@
 	health = maxHealth - getOxyLoss() - getToxLoss() - getCloneLoss() - total_burn - total_brute
 	staminaloss = total_stamina
 	update_stat()
-	if(((maxHealth - total_burn) < HEALTH_THRESHOLD_DEAD) && stat == DEAD )
-		become_husk("burn")
+	if(stat == DEAD)
+		if((maxHealth - total_burn) < HEALTH_THRESHOLD_DEAD)
+			become_husk("burn")
+		if(total_burn >= 400)
+			dust(drop_items = TRUE)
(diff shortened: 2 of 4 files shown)


UndeadNappist:
Narrowed the runtime issue to the fact gibbing calls to put blood on a weapon after the qdel happens.

Attempted to fix by re-arranging when it happened but, alas, didn't appear to work.

https://github.com/BadDeathclaw/TG-Claw/blob/13437b65b88a369451dba861e502f2819dfccd96/code/modules/mob/living/carbon/carbon_defense.dm#L66-L74

Appears to be the issue here. It seems to call itself attacking in a chain again, or something? I'm frankly stumped.

roboticxlink:
Ok, so, I found out the best way to get leather is by taking any sort of mutilated corpses around and gibbing them. Especially if there are few ghouls remaining or something. Or even if there still are ghouls. And so would this make it impossible to gib them? I mean, I guess it'd make sense that completely destroying the body would ruin the skin for leathermaking....

UndeadNappist:
> Ok, so, I found out the best way to get leather is by taking any sort of mutilated corpses around and gibbing them. Especially if there are few ghouls remaining or something. Or even if there still are ghouls. And so would this make it impossible to gib them? I mean, I guess it'd make sense that completely destroying the body would ruin the skin for leathermaking....

Destroying a body forces it to drop all contents using this code. I did _not_ want things being deleted off the body when I made this.

Similarly, this only affects /carbon/ not /simple_animal/ nor /silicon/, so NPCs, borgs, and robots, will be exempt.

roboticxlink:
To drop all contents? That's good at least.

Can you add code (maybe in this) to butcher body parts (arms, legs) and turn them into 1 meat each. And gibbing a body results maybe like, 5 meat? (or more meat per each body part)?

UndeadNappist:
Mobs are this weird thing where the limbs are totally unrelated to the body flying to pieces. If you manually varedit and delete every single one of a person's body parts, they don't die, and they don't fall apart.

Life and living are both VERY hackily coded and putting in butchering for the individual limbs - which are objects - is actually an entirely different chunk of code. Bodyparts are handled in a surgery module somewhere.

You could make it a separate PR and it would not interact with this in any way, shape, or form.

PR #257, opened 2018-12-03
Nerfs RnD point gain yo

ma44:
## Description

Reduces speed of point gain of RnD by 66%
## Motivation and Context

Requested by RenGusta
## How Has This Been Tested?

If a simple number change doesn't work then I'ma die


ma44:
>oh hey lemme merge all of this stuff right in here

PR #260, opened 2018-12-04
Nerfs shotguns

UndeadNappist:
-Removes captain's HK and replaces it with a 1911.

-Increases inaccuracy spread of pellets for shotguns, and decreases their damage.


UndeadNappist:
Conflicts fixed.

PR #268, opened 2018-12-06
Fixes up porta turrets

ma44:
## Description

To fix a issue related to the faction setting and targeting
## Motivation and Context

So you can have more faction based defenses
## How Has This Been Tested?

Not tested but if it compiles it should be better than before


ma44:
>tfw master contaminated

PR #288, opened 2018-12-17
Important: ignore dbconfig.txt

rhicora:
Avoid smushing server's dbconfig and admin file with the example ones in the repo.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,9 @@
 #Ignore byond config folder.
 /cfg/**/*
 
+#Ignore byond config folder.
+/config/dbconfig.txt
+
 #Ignore compiled files and other files generated during compilation.
 *.mdme
 *.dmb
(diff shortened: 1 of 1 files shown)


AndrewL97:
dbconfig and admins.txt are currently tracked (in the repo)
From what I gather, to have them actually be ignored, you will need to also delete them from the repo, so they become untracked.

This means that any future contributors have to manually create the admins.txt and dbconfig, additionally this will need to be done for server deployments.

The other option is to zip and delete the config folder, so people can just unzip.

https://git-scm.com/docs/git-reset
>--hard
Resets the index and working tree. Any changes to tracked files in the working tree since <commit> are discarded.

https://git-scm.com/docs/gitignore
> A gitignore file specifies intentionally untracked files that Git should ignore. Files already tracked by Git are not affected; see the NOTES below for details.

PR #293, opened 2018-12-19
Update observer.dm

UndeadNappist:
Basically only affects adminghosts in terms of usability, but observers will be able to view the UI of machines now.

Adminghosts will have full control over various.

### DIFF
--- a/code/modules/mob/dead/observer/observer.dm
+++ b/code/modules/mob/dead/observer/observer.dm
@@ -515,6 +515,9 @@ This is the proc mobs get to turn into a ghost. Forked from ghostize due to comp
 	set hidden = 1
 	to_chat(src, "<span class='danger'>You are dead! You have no mind to store memory!</span>")
 
+/mob/dead/observer/IsAdvancedToolUser()
+	return TRUE
+
 /mob/dead/observer/verb/toggle_ghostsee()
 	set name = "Toggle Ghost Vision"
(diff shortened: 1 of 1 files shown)


AndrewL97:
@LetterN @UndeadNappist
Has this been tested as:
Observer - Non-Admin
Observer - Admin
Ghost from Death - Non-Admin
Ghost from Death - Admin ?

Would need to check things like, standing on or beside the machine, being
away from it, etc.

Would hate to have normal ghosts poking things and being cheeky

AndrewL97:
Just checked @UndeadNappist 
Mins can access UI already, so this is really just adding it for non-admins
@LetterN do we want this ?

UndeadNappist:
As it turns out, ghosts cannot play the fiddle. Nor can they write.

Observers, ghosts, adminghosts and ghosting from bodies cannot use the UI (and couldn't before various updates; someone probably would have said something about ghosts tampering with the chemdispenser...).

Aghosts have problems with accessing the chemmaster and chemdispenser without these changes according to Ducky.

Post changes, all I can notice is all ghost types can see the UI, but _not_ change anything - which includes aghosting.

Perhaps there's a carryover from adminghosts, or an expectation that they could do something in another server(?)

LetterN:
@AndrewL97 as long as ghosts cant play pianos

PR #297, opened 2018-12-22
Adds locks to fence doors

UndeadNappist:
What it says on the tin.


UndeadNappist:
Works fine in the local test, screwed around with making mechs shove the lock open, wasn't particularly happy with it.

No worries, mechs can still bash down the fence in about four hits to break it.

PR #299, opened 2018-12-23
Fix for loot spawning outside of closets, crates

EvenInDeathIStillServe:
## Description
Crates and closets now wait for other objects to spawn before trying to move them to their contents.

## Motivation and Context
Makes it so that randomized loot spawns inside of containers instead of on top of them.

## How Has This Been Tested?
Randomized loot now spawns inside of containers properly.
Compared initialization times before and after the change, noticed no discrepancy.

### DIFF
--- a/code/game/objects/structures/crates_lockers/closets.dm
+++ b/code/game/objects/structures/crates_lockers/closets.dm
@@ -40,7 +40,8 @@
 	update_icon()
 	PopulateContents()
 	if(mapload && !opened) //now takes it first because timery thingies does not work with these
-		take_contents()
+		spawn(0)
+			take_contents()
 
 //USE THIS TO FILL IT, NOT INITIALIZE OR NEW
 /obj/structure/closet/proc/PopulateContents()


code/game/objects/structures/crates_lockers/closets.dm
```diff
@@ -40,7 +40,8 @@
 	update_icon()
 	PopulateContents()
 	if(mapload && !opened) //now takes it first because timery thingies does not work with these
-		take_contents()
+		spawn(0)
```
LetterN: sleep(0)

EvenInDeathIStillServe: What is the difference? I assume it performs the same purpose, but is it more optimized?

LetterN: sleep is apparently much better than spawn

EvenInDeathIStillServe: sleep(0) did nothing. spawn(0) is the way to go.

LetterN: alright then

Issue #265, opened 2018-12-06
Slot Machine Payout Bug


roboticxlink:
When you win with the slot machine X caps (400 is what I saw), it drops 400 caps, and also adds 400 caps to the machines credits, effectively giving you 800 caps.

#Feature not a bug. Totally.

GhostEcho:
Fixed by #301

PR #301, opened 2018-12-23
Slot machine payout fix

EvenInDeathIStillServe:
## Description
Removed some math that did the exact opposite of what should be done.

## Motivation and Context
The slot machine used to reward you twice when you get a payout: the first time by spawning caps, and the second time by giving the machine the same amount in its internal balance which can be withdrawn again.
Fix for #265 

## How Has This Been Tested?
Spent 10 minutes with 6 spawned slot machines until I got a payout. The slot machine did not get an extra payout in its internal balance.


rhicora:
@AndrewL97 can you review since you wrote this?

PR #303, opened 2018-12-24
It's the 21st century.

rhicora:
Updates snacks_meat.dm


rhicora:
It comes in the box of American food so there's no excuse.

PR #310, opened 2018-12-28
Adds gender reassignment

UndeadNappist:
This was probably a mistake. People wanted it, though.


code/modules/mob/living/carbon/human/examine.dm
```diff
@@ -103,6 +103,12 @@
 		if(100 to 200)
 			msg += "<span class='warning'>[t_He] [t_is] twitching ever so slightly.</span>\n"
 
+	if(gender_ambiguous) //someone fucked up a gender reassignment surgery
```
LetterN: it isn't fucked up, it was removed

PR #316, opened 2018-12-29
Fixes crosses

UndeadNappist:
-Makes it so unbuckling from crosses actually works.

-Makes it so they don't arbitrarily damage you just for being strung up on one.

-Makes it so you aren't rotated upside-down when you get buckled to a cross (Really?)

-Plays the correct noises now.

-Does the correct animations now.

-Reduces damage from being pulled and put onto a cross.

-Better flavortext

-Better, less buggy interactions

-Very minorly reduces damage from being on a kitchen spike so you don't just resist free and then instantly die of bloodloss.

-Cleans up some code.

There's probably a way to rewrite this chunk of code so there's less copypasting.

##
tl;dr the entire cross code was busted and not being used and it was defaulting to parent kitchenspike code

@rhicora

### DIFF
--- a/code/game/objects/structures/kitchen_spike.dm
+++ b/code/game/objects/structures/kitchen_spike.dm
@@ -76,26 +76,27 @@
 			L.forceMove(drop_location())
 			L.emote("scream")
 			L.add_splatter_floor()
-			L.adjustBruteLoss(30)
+			L.adjustBruteLoss(20)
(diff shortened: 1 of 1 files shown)


rhicora:
How tested is this?

UndeadNappist:
E x t e n s i v e l y.

PR #322, opened 2018-12-30
Adds EXP req to legion vet decanus

UndeadNappist:
Also hardcaps the # of possible spawn-in faction members. Recruit is twelve, legion recruit is 15.

No more infinity.

### DIFF
--- a/code/modules/jobs/job_types/legion.dm
+++ b/code/modules/jobs/job_types/legion.dm
@@ -114,6 +114,7 @@ Veteran Decan
 	spawn_positions = 1
 	supervisors = "the centurion"
 	selection_color = "#ffeeee"
+	exp_requirements = 1500
 
 	outfit = /datum/outfit/job/f13vetdecan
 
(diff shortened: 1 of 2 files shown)


rhicora:
Can we double check that Cent, Overseer, Captain, and Elder are gated behind experience reqs too?

UndeadNappist:
https://github.com/BadDeathclaw/TG-Claw/blob/295ea4f312ca39e9bcf76720126cb643bd0bece5/code/modules/jobs/job_types/bos.dm#L22

They're all gated, it was only vet. Decanus that wasn't. However, people are saying "all the command roles" which leads me to suspect something else is going on.

rhicora:
@UndeadNappist Rad, keep this pull updated with what you find ; )

UndeadNappist:
I believe this works, it'd be nice if it did.

rhicora:
Test it lolcatly!

UndeadNappist:
I did, used adminbussing to join in as a job, then later tried to re-join it after deadminning and wasn't able to.

However, I don't have a database set up.

PR #326, opened 2018-12-30
changes overseer time, has anyone realised i dont know how to add commits to existing PR's yet

companoo:
overseer was way too long


UndeadNappist:
Overseer has the highest potential for abuse, especially in such a tightly closed system. If we're having issues with having _no_ players for overseer, we'll consider it.

PR #348, opened 2019-01-04
Replace some passive research with active quests

rhicora:
This update gives the BoS something productive and interesting to do that fits with their lore.

- Adds the R&D Holotape, an item that can be destructively analyzed for 2500 research points, and scatters a bunch of these tapes around the wasteland
- Limits the ability to print more R&D servers (limiting passive research to 60/second, outside of scavenging more parts for another server)


AndrewL97:
@rhicora map diff bot sometimes just fails out of spite, its more or less just a quality of life thing to see the map changes, like icon diff bot. 
the only important one is pretty much Travis

JarlElbow:
I like this, though if I remember correctly, 2500 research points is only enough to get a single piece of minor technology, unless the values have changed. Perhaps raising that to like 5000 would be slightly more fitting, given that I imagine they would be slightly uncommon to come across?

DIamonnd:
Seems good enough. Although I believe it should either stay at 2500 or 3000. Any more would give the possibility of BoS getting ALL technologies. If they actually had to pick what is a priority depending the situation it would bring more importance to the Scribes work.

Cadwell-Regis:
Considering how loot spawns are going to become a lot more diverse and the focus on dungeons, it's likely many non faction groups will have the opportunity to do their own RND work in labs they've built themselves, so having lots of lootable disks with low point counts spread around can guarantee more often no single group gets access to them all.

GhostEcho:
Can I assume this isn't being worked on any longer?

PR #353, opened 2019-01-06
Get Rid of sec borgs

Monika4:
its a borg that can fire disabler with nearly no cooldown and have a stun baton and have a flash and on top of that it can be spammed if enough people is dead. Vault Makes it cancerous because they have the power to just spam sec borg

## Description
Most Annoying thing to fight againist and really abusable by the two tech powers aka BoS and Vault, Mainly Vault
## Motivation and Context
a borg that has rapid firing disabler gun and a stun baton as well as a cuffs so pretty powerful thus annoying because you either need to have EMPs or Ion Rifle (Ion carbine in this case)

### DIFF
--- a/config/game_options.txt
+++ b/config/game_options.txt
@@ -267,7 +267,7 @@ EVENTS_MIN_PLAYERS_MUL 1
 
 ## Secborg ###
 ## Uncomment to prevent the security cyborg module from being chosen
-#DISABLE_SECBORG
+DISABLE_SECBORG
 
 ## Peacekeeper Borg ###
 ## Uncomment to prevent the peacekeeper cyborg module from being chosen


JarlElbow:
Eh, I personally think just removing them isn't a perfect solution, instead of just having them nerfed. 
Since combat robots are a thing that both factions use, even if they currently are a bit OP with how quickly they can fire and stun.
However, security borgs are really the only reason the BoS would make a borg to begin with, and the vault rarely makes anything else, so removing them would practically be like removing borgs altogether for two factions.

Monika4:
its more disabling the Module than removing borgs as a whole

imw-Oculus:
Nerfing is preferable to removal. Some options.
- Increase the energy drain from attacks
- Reduce speed
- Increase cooldown between attacks

Monika4:
yeah I see now that it probly wouldn't be the best Idea but they hard to defeat i'll have a look for attack values or something

imw-Oculus:
Another option would be to reduce their maximum health/armor.

Monika4:
Yeah, that does sound like good ideas oculus

GhostEcho:
The issue is that borgs have no damage slowdown, so even if you reduce their health, they'll just yeet the fuck out of there and get repaired.

There's a reason why secborgs got removed from regular /tg/ and why very little people were upset.

GhostEcho:
Also, if secborgs are the only reason to make borgs, then that sounds like theres an issue with the usability of medical, engi and janiborgs, or the playerbase itself

Monika4:
personnel im sided with Main TG codes idea of yeeting them completey but still it may not be the best idea for this server

GhostEcho:
Really, its the best idea for any server in my opinion

imw-Oculus:
Adding damage slowdown is another option, yes. There are many, many options besides removing them completely.

GhostEcho:
Complete removal is the best though, because seriously secborgs disencourage the use of other modules, and add little to the fun of the game. Also bring back eyebots

JarlElbow:
Well, if we remove sec borgs, I'd like to add in some sort of combat borg/bot that we could use in place of that, since that sort of thing was always so prevalent in Fallout, especially when used by like the Brotherhood and other high-tech factions.

EndaruI:
the fact that they are spamed and out place they sure need a nerf or even removing them literaly some factions depend on this shit

Cadwell-Regis:
GUYS REMEMBER. RND Point Gain, Machines and Mining are being nerfed. So try not to double or triple nerf it just in case it gets fixed.

rhicora:
Sec borgs are a high-octane source of garbage and I've never seen one used in a manner that improves round quality. We'll increase Vault sec slots if we feel security needs are not being adequately met.

Approving this, no merge 'til Ren weighs in.

PR #363, opened 2019-01-13
Jet aint gonna make you Want jet

Monika4:
increases limit before you want jet

## Description
Jet is kind of broken at this point and you can take jet and never get rid of i

## Motivation and Context
Based on player feedback 

## How Has This Been Tested?
Going to complie and run this bad boy


imw-Oculus:
Instead of changing the addiction threshold, I recommend adding RNG into the addiction test to see if someone gets addicted.

Monika004:
that's the rand thing right?

imw-Oculus:
RNG = random number generator. So, yes. It makes more sense than just becoming addicted if you take too much of a substance.

PR #366, opened 2019-01-15
Makes PowerArmor unpiercable by shotgun darts/non piercing syringes

improvedname:
## Description
Makes powerarmor thick material, so it doesnt get pierced by shotgun darts/syringes

## Motivation and Context
Its a fucking a metal suit of protection, why would a silly syringe fit trough

## How Has This Been Tested?
Simple flag change.


imw-Oculus:
I'm sort of hesitant for this one, even though it logically makes sense. Could you possibly re-add syringe guns to map as part of this PR? Maybe have a box of piercing syringes as well?

improvedname:
Currently theres a argument, that the almighty power armor is shit, as its slow and you might aswell be a aim target, vault/bos can currently make piercing syrninge, and if raiders manage to raid the base, ((enclave base that is north of NCR) they can do too. even if it doesnt come much in play its just to prevent cheesy people using the beepsky smash/nuerotoxin drink to stun people trough shotgun darts/syringes, which can be a instadeath if you have a good mix

Monika004:
well it kind of falls back on the piercing thing being diamond headed thus having the force being put on a small area Aka how most heat rounds work but with out the copper lining shooting into a plate of armour its a tip piercing the armour

PR #370, opened 2019-01-16
Removes plasma cutters, from tech webs, and buffs normal mining drill speed to make it more appealing.

improvedname:
## Description
Removes plasma cutters from techwebs, nerfs the mining range trough rocks

## Motivation and Context
Plasma cutters are lame, and just easy run trough asteroid no effort at all giving BoS good equipment too fast.

## How Has This Been Tested?
simple value changes

### DIFF
--- a/code/modules/mining/equipment/mining_tools.dm
+++ b/code/modules/mining/equipment/mining_tools.dm
@@ -55,7 +55,7 @@
 	icon_state = "handdrill"
 	item_state = "jackhammer"
 	slot_flags = ITEM_SLOT_BELT
-	toolspeed = 0.6 //available from roundstart, faster than a pickaxe.
+	toolspeed = 0.4 //available from roundstart, faster than a pickaxe.
 	usesound = 'sound/weapons/drill.ogg'
 	hitsound = 'sound/weapons/drill.ogg'
 	desc = "An electric mining drill for the especially scrawny."
--- a/code/modules/projectiles/projectile/special/plasma.dm
+++ b/code/modules/projectiles/projectile/special/plasma.dm
@@ -3,12 +3,12 @@
 	icon_state = "plasmacutter"
 	damage_type = BRUTE
 	damage = 20
-	range = 4
+	range = 1
 	dismemberment = 20
 	impact_effect_type = /obj/effect/temp_visual/impact_effect/purple_laser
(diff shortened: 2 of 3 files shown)


DIamonnd:
Rather than this I'd see the RnD changes that had already been planned be a better solution. But what do I know.

improvedname:
put plasma cutter on dungeon map 👏

imw-Oculus:
Not sure about this one. I'm generally not in favor of removing features. Have you considered making it more expensive/require more plasma per use?

improvedname:
how about i make it use micro cells, and remove the plasma charge

Monika004:
pretty sure plasma cutters was getting removed in RND Passover 2.0

imw-Oculus:
@Monika004 Still retarded.

Monika004:
it do be like that chief, it do be like @imw-Oculus

Sanhaven:
A terrible solution to research, a blatant nerf to both Vault and BoS. Research already takes a lot of time to complete as is, and thats only if the team is half decent at best. It's a medRP server so your not going to get the best team all the time. Most rounds on med pop (47-50) last around 1-1.2 hours on average. Research is fully done around 50-60 minutes in, thats a full damn late game of league ffs. I don't know where everyone thinks BoS can finish research in less than 30 cause thats incredibly incorrect. The only time BoS shines is in actually the late game of the server, and the reason why they dominate is because NCR and Legion have either a.) shot each other, b.) shot themselves, c.) shot BoS while they don't have full strength anymore 1.5 hours into the round. 

There is a already a proposed change to research, but I'd also like to see plasma cutters see more widespead use- maybe add mining camps to where wasters, raiders, and settlers can use it so they can claim materials and use that as bargaining chips. There is so much potential with plasma cutters that the rest of the server can use- so instead of just dumbing down gameplay by nerfing everything ala Overwatch, I'd rather see an addition of content to other roles to provide more diversity. SS13 is unique due to the mechanics it naturally offers, all i've seen from admins these days are just blatant nerfs to quality, unique gameplay from SS13 that the rest of the server can benefit from.

JarlElbow:
You...do realize that all of this is already being added in the R&D update, right? 
Like, literally everything you put here is already being changed.

improvedname:
Be quicck

PR #371, opened 2019-01-16
Removes cloning from techwebs and replica pod seeds

improvedname:
## Description
Removes cloning from techwebs, and replica seeds from the vendor

## Motivation and Context
Litteraly gives dead no meaning, and you get some retarded shit like a BoS iniatate saying, LOL IF YOU KILL ME I GET BACK ANYWAYS

## How Has This Been Tested?
simple removal.

### DIFF
--- a/code/modules/research/techweb/all_nodes.dm
+++ b/code/modules/research/techweb/all_nodes.dm
@@ -396,15 +396,6 @@
 	export_price = 5000
 
 ////////////////////////Medical////////////////////////
-/datum/techweb_node/cloning
-	id = "cloning"
-	display_name = "Genetic Engineering"
-	description = "We have the technology to make him."
-	prereq_ids = list("biotech")
-	design_ids = list("clonecontrol", "clonepod", "clonescanner", "scan_console", "cloning_disk")
-	research_costs = list(TECHWEB_POINT_TYPE_GENERIC = 2500)
-	export_price = 5000
-
 /datum/techweb_node/cryotech
 	id = "cryotech"
 	display_name = "Cryostasis Technology"
--- a/code/modules/vending/megaseed.dm
+++ b/code/modules/vending/megaseed.dm
@@ -28,7 +28,6 @@
 					/obj/item/seeds/potato = 3,
 					/obj/item/seeds/poppy = 3,
 					/obj/item/seeds/pumpkin = 3,
-					/obj/item/seeds/replicapod = 3,
 					/obj/item/seeds/wheat/rice = 3,
 					/obj/item/seeds/soya = 3,
 					/obj/item/seeds/sunflower = 3,


DIamonnd:
1.- Only BoS ever makes them, even so, the braindeads never scan themselves.

2.- If the initiate says that, they are a shitter. They should be killed again.

improvedname:
YEaH oNly BoS cAN mAKe ThEM With HunDREdS oF mIniBOMbS anD aSSaULt RifLES so THeY Can MEtaRush YOU and KIll you In RevENge OMegAlol

bigfoun:
Cloning makes it so death has absolutely no meaning whatsoever. The BOS are all about self preservation and survival. This conflict makes their faction infinitely more interesting. Cloning takes away all worry of death because they know they'll just be coming back. In the case of the vault, it adds onto their already LRP behavior.

JarlElbow:
A lot of people on the staff team always say that you should try to improve something instead of just removing it if you don't like it.

Instead of just outright removing cloning, why not try to find an alternative solution? After all, RnD is being massively reworked in the coming week or so already, so it's going to be more difficult to research stuff like that over the course of a round. How about making it so that cloning has a higher chance of giving you a major defect that takes time and effort to heal? 

On top of this, there seems to be a trend of people wanting to remove stuff just because the BoS has it and they think the Brotherhood is filled with shitters. If we decided to nerf a faction at the first complaint we got, we'd turn into CM.

improvedname:
idk dude half of the staff team say something else about cloning 👏

improvedname:
besides science mains are always shitters lmfao 👏

Sanhaven:
Cloning is fine only to the extent of the individual and the team strength at the time. I feel cloning should not be removed, but regulated more a la rules (new life rule.) It should be made more of a staff thing IMO, but since you are targeting BoS directly, i'll give a case in your defense. 

Yes, I have seen initiates run back to RnD and begin to meta their old life back- if thats the case, then I aHelp it immediately whenever I see it. This is true. However is rare. Again, this should be handled staff wise as well, and should be set some cloning rules instead of chopping out content after content to make the game more braindead and non-fun. I personally tell Inits when I play to don't recall their past life, and admins play BoS frequently and also assist in that matter.

As for the defense of the "Litteraly gives dead no meaning" for BoS, every single death has a great meaning. With only SIX total members at one time, initiates being the only non-time-gated role; If a BoS is dead, it literally costs them more than a body; yes we have cloning, but the price is higher than anyone realises: 

1.) We can't respawn. Yes we have cloning, but most inits/BoS don't scan again, so death is absolute perma for the round. Unless an admin opens it up, Initiate is the only unique "grunt" rank that cannot respawn.

2.) ID and Comms are extremely compromising to the base's integrity. If an Inititate is killed, it means a major fuckup has occured, and now both comms and ID is taken, already putting a small team even smaller. BoS are the smallest faction in the game, so losing these two pieces of equipment is incredibly taxing as there is no known way to replicate it without admin intervention. Base would be wide open, and their comms (which is valuable) is taken. Raiders and Wasters have been known to meta the base, but even that opinion aside, it leaves NCR/Legion/wasters free entry to the base, and its not hard to robust a weakened BoS. One man really makes the difference.

3.) Dumb ass initiates: Are common, due to time gates. My advice is to simply aHelp them. 

In conclusion, again, instead of dumbing down content, i'd rather see an addition to the wastelands; or development for other roles to make them up to par with the BoS. Cloning should be handled better via staff intervention as well, and just aHelp those fucking idiot Inits please.

skubl:
Of the many things in BD that remove the player character's proclivity to want to stay alive and avoid danger, cloning appears to be one of the bigger contributors.

JarlElbow's comment about improvement over removal is an interesting one, but unfortunately, I think if cloning was more difficult or gave clones a disability/other crippling defect - as opposed to the faultless clones we get now - the average player may just avoid doing it altogether. 

It could end up being a lot of work to implement defects or whatever, only to achieve the same result as if cloning had just been removed.

imw-Oculus:
I'm going to leave this one up for a few days, and check with Ren to see what he thinks.

GoodG1448:
> 
> 2.) ID and Comms are extremely compromising to the base's integrity. If an Inititate is killed, it means a major fuckup has occured, and now both comms and ID is taken, already putting a small team even smaller. BoS are the smallest faction in the game, so losing these two pieces of equipment is incredibly taxing as there is no known way to replicate it without admin intervention. Base would be wide open, and their comms (which is valuable) is taken. Raiders and Wasters have been known to meta the base, but even that opinion aside, it leaves NCR/Legion/wasters free entry to the base, and its not hard to robust a weakened BoS. One man really makes the difference.
> 

There are plenty of ways to avoid item looting from BoS if you are creative enough. Lorewise, BoS is wary of letting other people get high tech shit, maybe not to the point fanatism of setting themselves on fire to burn everything they have with them, but still, it is a solution. CFL3 dental patches can do that, aswell deleting IDs and headsets in the process. Or maybe health sensor grenades triggered by the death of the carrier (This option is less dynamic since the user can't control its death, you can doing the CFL3 patch thing).

About the thing in general. I agree with Jarl about removing stuff being a shitty move. Looking at #348 gives me a clear answer that it is already coded into the game.

In regular /tg/, the only way to unlock the alien surgery tech tree is to get a alien tool and reverse engineer using the Destructive Analyzer. We could do the same with a lootable disk that spawn somewhere in the wastes for the Brotherhood (Or any waster willing to trade) to find and unlock the cloning tech tree.

improvedname:
Theres litteraly a surgery to revive people.

Idds can be remade by circuitry

PR #372, opened 2019-01-17
NCR Base Remap

JTGSZ:
## Description
NCR Base Rework
-Moved South and West, remapped
-Active turfs cleared

## Motivation and Context
It saves the power plant, adds more things to do at the NCR base that are non-combat related, resembles a municipality courthouse. There are three buildings part of the compound which add more defense and make assaults a bit more interesting, they are all themed too. Basic gardening of the regular vegetable variety is there too.

Northern area may resemble ancient NCR base design similar to power plant in openness, without being the same.

## How Has This Been Tested?
Tested on a hosted server


JarlElbow:
Well, Ren gave this his seal of approval, and it looks great, so +100 from me.

Issue #266, opened 2018-12-06
Wood Wall Deconstruction


roboticxlink:
The default wood walls (the beaten down white washed ones on most wasteland buildings) is supposed to be a wood wall, but the deconstruction (and probably stats for it) yields metal. It should be fixed to be a subtype of wood walls or something so they drop wood when deconstructed instead of metal.

PR #380, opened 2019-01-18
Makes Wooden Walls Wooden

GhostEcho:
Fixes https://github.com/BadDeathclaw/TG-Claw/issues/266, wooden wall structures on buildings actually spit out wood, instead of metal 

Credits to Oculus for helping guide me through this


GhostEcho:
Okay so basically runtimes out the ass *insert works on my machine here*

PR #381, opened 2019-01-18
Wooden Wall Fixes v2: Electric Runtime Edition

### LINKED ISSUE #266: Wood Wall Deconstruction

GhostEcho:
Fixes #266 so that starting wooden walls for appropriate buildings are made with wood, and are deconstructed with crowbars, leaving a wooden barricade underneath

Deconstruction gives 2 wood planks due to the rotting nature of the buildings

Hopefully no runtimes this time

Co-Authored-By: imw-Oculus


imw-Oculus:
This works just fine. It does spit a run-time relating to revs, but I don't think it was the fault of this PR. Investigate it a little, test with a blank master, and then get back to me about the origin of the runtime.

rhicora:
This doesn't pass map diff, so you'll probably need to rerun it.

GhostEcho:
alright, gonna close and reopen then

PR #392, opened 2019-01-21
Changes access of lethal injection locker to BoS, and removes access from anesthetics locker

improvedname:
## Description
For the changes made by the BoS remap


imw-Oculus:
Downloading and testing now.

PR #396, opened 2019-01-22
Nerfs BHS-M1 Melee, Buffs Ithaca, Winchester and Remingtons

GhostEcho:
Drops the BHS-M1 melee from 56 to 40 brute, changing it from 2 hits to 3 to crit an unarmored player

Buffs the Ithaca/Winchester/Remingtons from 10 to 25 melee as requested by Ren, changing it from 10 hits to 4 to crit an unarmored person


GhostEcho:
Ended up doing this in my pipe rifle fix as well, so closing just for management sake

PR #399, opened 2019-01-22
Makes Pipe Rifles and Zip Guns Reload with 10mm

GhostEcho:
Previously, pipe rifles and zip guns would spawn with a 10mm round in their chambers, but then be reloaded with shotgun shells due to their internal mags using shotgun code

Now they use revolver code instead with one ammo capacity, no more dual wielding zipguns with buckshot


rhicora:
Are you able to run the test server locally?

PR #414, opened 2019-01-23
Fixes the power armor hardsuits

ma44:
## Description
Fixes power armor by fixing some edge case
## Motivation and Context

so we can have power armor hardsuits
## How Has This Been Tested?

Tested some long time ago; chances are it works but ill keep it admin spawn for now only


UndeadNappist:
Should probably push all the power armor stuff in a single branch, and could you get someone to test it before merging?

PR #415, opened 2019-01-23
Fixes the porta turrets custom target by faction

ma44:
## Description

Fixes the porta turret custom faction adding by making it not add neutral faction tag
## Motivation and Context

To make automatic defense easier
## How Has This Been Tested?

By seeing that it targets people without the correct faction tag properly


rhicora:
Looks fine I guess.

PR #419, opened 2019-01-24
Bo Staff Changes

GhostEcho:
Ren allowed these changes
Block chance dropped from 50 to 30
Force from 25 to 30
Stamina dam on disarm intent instead of 13/20 rng is now a flat 19


GhostEcho:
I won't be entirely upset if I can't get travis working

GhostEcho:
Oh goddamnit Travis worked this time

GhostEcho:
Fixed, unfortunately

PR #424, opened 2019-01-25
Removes mining out entire mountains easily

GhostEcho:
Ranges severely reduced on the cutters, and they require two hands to fire instead of one.

Range change requested by Ren, two handed suggested by ImprovedName

### DIFF
--- a/code/modules/projectiles/guns/energy/special.dm
+++ b/code/modules/projectiles/guns/energy/special.dm
@@ -124,6 +124,7 @@
 	force = 12
 	sharpness = IS_SHARP
 	can_charge = 0
+	weapon_weight = WEAPON_HEAVY
 
 	heat = 3800
 	usesound = list('sound/items/welder.ogg', 'sound/items/welder2.ogg')
--- a/code/modules/projectiles/projectile/special/plasma.dm
+++ b/code/modules/projectiles/projectile/special/plasma.dm
@@ -3,12 +3,12 @@
 	icon_state = "plasmacutter"
 	damage_type = BRUTE
 	damage = 20
-	range = 4
+	range = 2
 	dismemberment = 20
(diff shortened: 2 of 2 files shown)


SpaceVampire:
Yes lets make one of the more tedious mechanics in the game take even longer to do. Mining is already anti-fun in its current state as unlike normal SS13 there are no side activities (Monsters, Ruins, Etc) and no opportunities for any type of human interaction.

GhostEcho:
to be quite honest, this change doesn't even make much of an impact, as someone who mained miner on /tg/ for over half a year, Id rather take out cutters completely but Ren said to go with a range nerf, the issue is that you can just keep spam clicking and mine faster than any other mining tool. But yes, feel free to kneejerk to a PR without actually trying the changes out.

GhostEcho:
All it does is make it so you have to reload more often

Cadwell-Regis:
Anyone who complains about mining hasn't played Foxhole, where people spend 2-3 hours non stop mining scrap fields for tech parts that make almost no difference to the war effort, but is an essential component anyway.

The mining in F013 alternatively is extremely quick, and with the materials you can supply enough gear and ammo for a fairly large team of super soldiers.

JarlElbow:
I'll be honest. People know that I hate big nerfs more than anything. 

This isn't that bad. The plasma cutter is so unbelievably fast right now, and having its range reduced isn't that bad in the slightest. If I remember correctly, it was lowered to 2, which still means you can mine faster than any drill you could make, while still being slow enough to avoid early game rushes from certain factions.

improvedname:
people screaming at the convience removal for the sake of a little effort

imw-Oculus:
Lots of convo on this one, going to wait a bit for additional input before merging.

PR #429, opened 2019-01-28
Allows legionnaires to use instruments

GhostEcho:
Im so sorry it's taken this long Legion.


rhicora:
Sexy Sax Vex enabled.

PR #430, opened 2019-01-28
Bluespace Trauma Acquistion Removal [Do Not Merge Yet]

GhostEcho:
AAAAAAAAToday at 3:27 AM
or the bluespace wormhole trauma
wherein
you can use pocket dimensions
to slip around the map

AAAAAAAAToday at 3:27 AM
it is very handy, especially when it lets you slip into the comms room in the far right of the map
which is not immune to that trauma
and you break the comms machines

This should explain why


AdamElTablawy:
considering i've used this to repeatedly get onto the centcomm zlevel, take the pulse rifles or fly the syndicate shuttle around

this needs to go

PR #431, opened 2019-01-28
Saves us all from this degenerate hell hole

GhostEcho:
purges world.dm


GhostEcho:
The worst part of this meme is that it actually compiled

PR #432, opened 2019-01-28
Legion Speedy Bois Foundation

GhostEcho:
adds in trait for eventual legion speed buff, doesn't give them the trait


UndeadNappist:
Altering player speed is a horrifically and drastic alteration to game balance. I would strongly recommend against even having this.

I'd rather tackle how guns are balanced versus melee weapons first, along with creating greater mechanical encouragement for people to kidnap each other instead of murder.

PR #434, opened 2019-01-28
removes cloning (Idk if this will work)_

companoo:
how do i test shit guys

### DIFF
--- a/code/modules/research/techweb/all_nodes.dm
+++ b/code/modules/research/techweb/all_nodes.dm
@@ -396,14 +396,6 @@
 	export_price = 5000
 
 ////////////////////////Medical////////////////////////
-/datum/techweb_node/cloning
-	id = "cloning"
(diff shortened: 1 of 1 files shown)


improvedname:
doesn't matter, it auctally functions in fact 👏  now speedmerge

companoo:
i actually figured out how to test shit and can confirm this removes cloning

UndeadNappist:
This just removes it from the techweb node, not the circuit board or a bunch of other stuff from mapping or other discreet parts of code.

This also leaves the alternative methods of revival.

If you want to update cloning, I would recommend replacing all acquirable cloning devices with experimental cloning instead ; it grabs a random ghost for your new clone instead of the original and, I believe, can also be used to make multiple clones of the same person.

Leave cloning in for now as it's plugged into a bunch of other stuff unless you want to do a very thorough cleanup.

Issue #417, opened 2019-01-23
Brotherhood Scribe ID doesnt have cyborg access.


SpaceVampire:
Issue reported from  Round ID: 768

Reporting client version: 512

Unable to lock or open cyborgs.

PR #436, opened 2019-01-29
Gives Brotherhood robotics access

companoo:
issue #417


GhostEcho:
Only issue with this is that they'd be able to open vault robots too which is why I hadn't done it before

PR #438, opened 2019-01-29
legion armory changes

companoo:
adds caravan shotguns
adds more shields
removes two boxes of single shotguns one still remains
adds more spear quivers


skateboardjones:
yes

PR #443, opened 2019-01-30
Hotfixes runtiming walls

UndeadNappist:
Correct fix is re-writing girder and construction code. This will do for now.

Notably, there's a conflict between trying to call on contents, closed/open air turfs, and the fact that there's a conflict between it trying to figure out whether it's got a girder inside of it or a barricade. For now this just sidesteps the issue, but allows people to convert wooden barricades into metal girders by building on them then deconstructing the wall.


rhicora:
Needs to pass build before merging though.

UndeadNappist:
Issue fixed.

PR #448, opened 2019-01-30
wastebots will attack you while in critical condition

AdamElTablawy:
you know, like goliaths do, so that you can't cheese the enclave bunker solo anymore.

why: because i can cheese the enclave bunker solo with healing meds in less than an hour from roundstart.

what this means is if a gutsy, protectron, or handy crits you, it will keep hitting you until your lungs stop doing the take in air in and out movement and brain function turns off.


UndeadNappist:
Protectrons charging you was to address an issue where, while they moved at mach 5, could clean up everyone.

Their speed's two so that's a fine change as of now.

Could you also change the melee damage of Gutsies to flame and give them a description of 'burns you with their flamethrower' while you're at it? As it is in their description.

PR #455, opened 2019-01-30
Buffs Ammo/Grenade Pouch

Raze500:


Raze500:
> As an heirloom, wouldn't you become immediately unhappy when it leaves your inventory because you used it or it got destroyed?

Already tested it and no, Don't gives unhappy mood/happy mood. also has you can see the diferences... the original heirloom have the adding/removing mood variables.

I've removed them to make it work just like an normal item.

PR #456, opened 2019-01-30
Hotfix for PDAs

UndeadNappist:
As they're inserted into the glove slot, the game was constantly checking to see if they left fingerprints whenever you touched anything.

It didn't have var/transfer_prints so it was spewing runtimes instead.

Be more careful.


kilozombie:
Not a real coder but couldn't you put that var in the parent type so this never happens again with new gloves?

PR #462, opened 2019-01-31
Update job.dm

UndeadNappist:
As stated in comments.


rhicora:
This is acceptable code but should be discussed before merging. @RenGusta

RenGusta:
If these are the job hierarchy changes, and they don't agree over things like the timelocks system, should be A-Okay.

Issue #362, opened 2019-01-12
Let Pushers spawn with ANYTHING other than the Jabroni outfit


Monika4:
Was Requested to not have Kinky Drugos

PR #463, opened 2019-01-31
Gives pushers more uniforms to pick from.

improvedname:
## Description
Lets it pick from more uniforms

## Motivation and Context
#362
No more kinky druglords in a jabroni outift.

## How Has This Been Tested?
works


rhicora:
I think this is fine to merge. We can throw some jabronis in the back of the building they spawn in if that's the concept they want to go for.

PR #466, opened 2019-01-31
Nerfs Shotgun Darts

GhostEcho:
30 units down to 15 units, no more instant death mix by mixing 3 healing powder together in it


skateboardjones:
why don't you just remove them completely

PR #468, opened 2019-01-31
Removes Prescanning, autocloning

JTGSZ:
Removes Pre-scanning, Autocloning.

## Description
Just removes the ability to Pre-scan and Autoclone

## Motivation and Context
While watching BD yesterday, I witnessed a BoS member using the prescanning and autocloning to literally play what amounts to a FPS game with the wasteland, as everytime he'd die. He would get automatically cloned taking nobody with him each time despite the fact BoS is supposed to be tightknit, then reequip and march out to antagonize and harass wasters again. Whom have no form of auto-respawn.

That basically is showing a deeper problem in the existence of autocloning, and prescanning since that is how it functions at a basis on all stations. I used to enjoy trying to salvage brains and bodies since I felt like a saint, or as an antagonist knowing my target was gone If I disposed of the corpse. Not to mention being told to get scanned as a crew? It feels bad, knowing I will come back If I die with nobody around detracting from my gameplay experience, and knowing that I'm being pushed to do so coddling me. This is something I've always never been a fan of as a longtime player.

## How Has This Been Tested?
Hosted and Tested

alive
untitled

### DIFF
--- a/code/game/machinery/computer/cloning.dm
+++ b/code/game/machinery/computer/cloning.dm
@@ -6,18 +6,13 @@
 	icon_screen = "dna"
 	icon_keyboard = "med_key"
 	circuit = /obj/item/circuitboard/computer/cloning
-	req_access = list(ACCESS_HEADS) //ONLY USED FOR RECORD DELETION RIGHT NOW.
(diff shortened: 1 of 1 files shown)


Raze500:
Agree, i was the westlander.

PR #495, opened 2019-02-02
Prosthetics Overhaul 

JohnOxford:
## Description
adds different tiers of prosthetic limbs, for upgrades and the such in preparation for modular limbs.
tier 1 - surplus limbs
tier 2 - salvaged limbs
tier 3 - forged limbs
tier 4 - manufactured limbs
tier 5 - alloy limbs
tier 6 - combat limbs

## Motivation and Context
in the future, adding a negative quirk for getting a high tier limb would be a thing, also allows admins
to spawn them in in the future, additionally, map changes would allow advanced limbs to spawn in
dungeons

only tier 6 limbs are "better" than normal human limbs, the logic being that if you have a amputation, having a limb slightly worse than normal human limbs would be better than having a limb that is significantly worse. which all in all is better than having no limb at all.

future plans including adding quirks for it, not requiring surgery to switch limbs, and the ability to repair or upgrade limbs (perhaps giving all the useless shitty items laying around a purpose)

## How Has This Been Tested?
Travis and such

### DIFF
--- a/code/modules/surgery/bodyparts/robot_bodyparts.dm
+++ b/code/modules/surgery/bodyparts/robot_bodyparts.dm
@@ -216,31 +216,163 @@
 	..()
 
 
-
+// tier 1
 
 /obj/item/bodypart/l_arm/robot/surplus
-	name = "surplus prosthetic left arm"
+	name = "surplus mechanical left arm"
 	desc = "A skeletal, robotic limb. Outdated and fragile, but it's still better than nothing."
 	icon = 'icons/mob/augmentation/surplus_augments.dmi'
 	max_damage = 20
 
 /obj/item/bodypart/r_arm/robot/surplus
-	name = "surplus prosthetic right arm"
+	name = "surplus mechanical right arm"
 	desc = "A skeletal, robotic limb. Outdated and fragile, but it's still better than nothing."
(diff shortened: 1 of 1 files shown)


UndeadNappist:
Do each of these new limbtypes have a different in-hand and different on-body sprite?

JTGSZ:
Im curious to see what kind of icons you drum up.

JohnOxford:
they're planned in the foreseeable future, but with the limited amount of pixels for individual limbs, i wouldn't expect anything ground breaking. perhaps both ends of the spectrum (tier 1 and tier 6) will resemble the curved bladed spring running thing, short of that, everything in between would likely end up being just different varieties of grey/black/white combos. maybe adding some lighting for the upper tiers.

RenGusta:
This isn't at all appropriate for our server. If you could implement a single kind of prosthesis, being 'improvised' that works as a dummy limb with no actual use, like a wooden arm to let you use two handed weapons if you lose a limb, that would be all I support.

We don't need 6 tiers of prostheses when we would barely have more than peg legs.

UndeadNappist:
I just wanted them to be visually different and interesting. Part of the game is visual.

PR #500, opened 2019-02-03
Revert "Ma fix your goddamn code"

improvedname:
Reverts BadDeathclaw/TG-Claw#484


UndeadNappist:
Needs to be connected with a fix that makes it less likely to cause the same runtime and oopsy issues again. Put this in front of another commit for the armor itself.

PR #503, opened 2019-02-03
updates jet recipe so it doesent need a spraycan

companoo:
spray cans are rare and people should make more jet

also adds welder to make it harder(?)


Kimochiiiiiiiiiiiiii:
You know you can print them in autolathes and that they are cheap to, right?

UndeadNappist:
Why not make empty spraycans hand-craftable from steel sheets? Then we can add mechanics for filling it and making paint later.

PR #522, opened 2019-02-05
Map Changes, Bugfixes

JTGSZ:
## Description
Cleaned up Area Icons.
Added Xenobio to vault.
Fixed map errors.
Reduced Lag

## Motivation and Context
BoS gained xenobio, so the vault did too.
Area Icons were too opaque making it a pain in the ass sometimes.
Minor things that were off were fixed.

## How Has This Been Tested?
Hosted and tested.


JTGSZ:
> Remove timestop ability from sepias, as well as sentience transfer potions from rainbows, I'd also recommend removing angel wings from them as well, but that doesn't bother me so much, gold slime hostile mobs could be an issue but everyone has guns or can get them easy in the vault, so it should be okay

if you pay more attention, you would realize slime_extracts.dm was already changed.

GhostEcho:
Then never mind, wasn't aware that got done already

PR #540, opened 2019-02-09
Adds greyscale dildos

improvedname:
dildos now spawn with random colors


Raze500:
**HmmMmmmMMm Yes finally i can't harrass people with differents colours.**

PR #546, opened 2019-02-11
Attempt numero uno on jobs and quirks not saving

UndeadNappist:
Fixes quirk preferences being carried over, but isn't a real 'fix'.

Job preferences now save properly.

What's happening is that the datum is being saved somewhere in the interim state and then instead of being cleared and overwritten by selected slots, it's overwriting selected slots regardless of what they are.

Navigating this horrible maze is giving me a headache, so I'll pick this up later.

### DIFF
--- a/code/__DEFINES/jobs.dm
+++ b/code/__DEFINES/jobs.dm
@@ -42,6 +42,8 @@
 #define MIME			(1<<12)
 #define ASSISTANT		(1<<13)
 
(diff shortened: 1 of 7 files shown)


UndeadNappist:
Fixed now, no runtimes. Saves properly.

It could probably be optimized but for now you get spaghetti.

UndeadNappist:
Tests done, ready to merge. @rhicora

rhicora:
I trust you honey.

PR #550, opened 2019-02-11
Update legion.dm

UndeadNappist:
Same problem with Legate, only working the other direction. Job is still joinable, just not selectable on setup.


Raze500:
Wait so you can legate now?

PR #561, opened 2019-02-13
Holopad design removal

GhostEcho:
Due to increasing amounts of "not" metagame by the BoS using it to "not" bypass the spirit of the no metagaming rules to find the vault, holopads are being removed from RnD designs. Will potentially be re-added in if we can create separate networks for them, but for now, cutting down the amount of "not" metagaming


ErwinSnowden:
Add faction protolathes, Vault can print most things except security/defense technology(the state of their protolathe is abysmal ever since the nerfs, can't even make a fucking mining drill), Enclave can print anything with their RnD set and the BoS can print whatever fits their bill i guess. Vault still uses them and it's a pain in the ass to not be able to build more for announcements on holodisks, We could put a holopad outside of the vault door so people could contact us with it, put one into the overseers office too, with a limited amount of holopads for every faction what am I supposed to do? Intercomms are broke.

improvedname:
erwin does has a point. its time to kill the omnilathe off

improvedname:
@UndeadNappist this can be closed, gonna seperate the bos/vault tec network in the materials rework.

GhostEcho:
@improvedname good man then

PR #566, opened 2019-02-14
updates cigarettes to be not spacey + techweb shit i have no idea how to fix

companoo:
title tbh

reason i deleted most cigs is because they arent ingame, we can add our own later if needed


UndeadNappist:
Try restarting from a cleaned up master branch and then putting the changes you want in there.
1. Copy your changes into a separate file somewhere and save it.
2. Open the command prompt.
3. git checkout master
4. git pull https://github.com/BadDeathclaw/TG-Claw.git master
5. git rebase origin master -force
This'll make your master branch look exactly like this git's master branch, and you'll lose everything in it so be careful.
Then do the edits in a new branch from there.

PR #568, opened 2019-02-14
cigarettes this one actually works

companoo:
updates cigarettes to real life variants, basically QOL

doesn't change there hierarchies


rhicora:
This breaks the build, please look into why that is happening.

UndeadNappist:
But it does delete paths or change them, which means that things that may call on them in a list will break. Check what Travis says.
Also remember to compile before submitting.
image

companoo:
not sure how i fucked up cigs

companoo:
fixed the one error that came up

UndeadNappist:
Did you double-check to investigate how the vending machines call on lists so it doesn't runtime every time you try to interact with or spawn a cig vending machine?

PR #569, opened 2019-02-14
removes space from instruments

companoo:
removes a few words

### DIFF
--- a/code/game/objects/items/devices/instruments.dm
+++ b/code/game/objects/items/devices/instruments.dm
@@ -44,7 +44,7 @@
 
 /obj/item/instrument/violin
 	name = "violin"
-	desc = "A wooden musical instrument with four strings and a bow. ""
+	desc = "A wooden musical instrument with four strings and a bow."
 	icon_state = "violin"
 	item_state = "violin"
 	hitsound = "swing_hit"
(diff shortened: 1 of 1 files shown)


rhicora:
This breaks the build, please look into why that is happening.

code/game/objects/items/devices/instruments.dm
```diff
@@ -43,16 +43,16 @@
 	song.interact(user)
 
 /obj/item/instrument/violin
-	name = "space violin"
-	desc = "A wooden musical instrument with four strings and a bow. \"The devil went down to space, he was looking for an assistant to grief.\""
+	name = "violin"
+	desc = "A wooden musical instrument with four strings and a bow. ""
```
UndeadNappist: has an extra space and two quotation marks
You can also just change the reference to the original lyrics.

code/game/objects/items/devices/instruments.dm
```diff
@@ -43,16 +43,16 @@
...
 	icon_state = "violin"
 	item_state = "violin"
 	hitsound = "swing_hit"
 	instrumentId = "violin"
 
 /obj/item/instrument/violin/golden
 	name = "golden violin"
-	desc = "A golden musical instrument with four strings and a bow. \"The devil went down to space, he was looking for an assistant to grief.\""
+	desc = "A golden musical instrument with four strings and a bow.""
```
UndeadNappist: Same as previous issue, only no extra space, just an extra quotation mark.

Issue #383, opened 2019-01-18
Scoped weapons can use their scope in belt and suit slots


GhostEcho:
Super cheesy, allows people to snipe with weapons they shouldn't be able o snipe with due to a lack of scope

Also explain to me how the fuck this would work

Not I dead pls nerf, moreso I saw and was disgusted with it

UndeadNappist:
if(!in_hand)
return

Something like that, anyway.

Move all zooming proc code out of the individual guns and weaponry and move it into /obj/item in zoom.dm, ensure the guns/binocs that zoom call on it and only that.

Step two is bugfixing the hell out of zooming. Recommend finding the least-broken zooming device and stamping all the others to use that code and tinkering from there (a harder chore).

JTGSZ:
Solved by #577

PR #577, opened 2019-02-15
Scope Fixes

JTGSZ:
## Description
You can't scope from your pockets or other slots except hand with anything that is zoomable now. 

Solves Issue #383 

## Motivation and Context
Scoping from pockets and suit slots is now not a thing.

## How Has This Been Tested?
Hosted and Tested

### DIFF
--- a/code/modules/projectiles/gun.dm
+++ b/code/modules/projectiles/gun.dm
@@ -92,8 +92,9 @@
 
 /obj/item/gun/equipped(mob/living/user, slot)
 	. = ..()
-	if(zoomed && user.get_active_held_item() != src)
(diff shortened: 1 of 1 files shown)


rhicora:
359 no scope in your goddamn pocket.

UndeadNappist:
It feels bad to put binocs and scopes in the same folder, and also not just give them the same line of code that's then called ZOOM or something, but this works.

UndeadNappist:
https://github.com/BadDeathclaw/TG-Claw/blob/002e22162550f494011093ce1c3ea2943149e172/code/modules/projectiles/gun.dm#L486

Issue #524, opened 2019-02-06
Vault Doc Access


GhostEcho:
they cannot use the surgery shutters button, nor the chem desk windoor apparently

JTGSZ:
Solved by #578

PR #578, opened 2019-02-15
Map Fixes, Changes

JTGSZ:
## Description
Added some Surgery Disks like requested to the hospital.
Cleaned up some some off turfs and objects.
Fixes Vault Surgery Shutter Button being incorrect (#524)
Fixes Vault Chemistry Desk Door Access being Set to 50 instead of 33. (#524)

## Motivation and Context
Some turfs were off like the tiles under the diner windows, or the single interior door in vault security in the middle of it, also found some missing wires in the vault.
Solves Issue #524 

## How Has This Been Tested?
Hosted and Tested.


rhicora:
Can this fix the fence issue in Legion base?

JTGSZ:
I can handle it real fast

Issue #195, opened 2018-11-16
Wanted posters spawn with no target info and can't post them up


roboticxlink:
[Priority]: Low

[PR]: Fixing loot spawns for the wanted posters.

[Reproduction]: Any number of Wanted Poster () loot spawns

[For Admins]: 

[Notes]:

RedTekRedd:
Will investigate.

JTGSZ:
Solved by #579

PR #579, opened 2019-02-15
Replaces bugged Undefined Wanted Poster that is in Junkspawner

JTGSZ:
## Description
Replaces Wanted poster that needs security records to define with a random contraband poster, adds random official poster in place of cigarbutt too. Solves Issue #195 

## Motivation and Context
Wanted poster can't even be used on the wall, and just disappears. It needs security records to function and cannot be spawned blank which it does in the junkspawner.

## How Has This Been Tested?
Hosted and Tested with a hundred Junkspawners


rhicora:
Sure!

PR #584, opened 2019-02-16
You can no longer move at permanent ephedrinespeed through creating oldschool nuka cola

AdamElTablawy:
Basically, due to something nobody bothered to address, you could move at extremely high speeds by mixing regular nuka cola-uranium and cola. 

Uranium is everywhere in the wastes and cola is available quite commonly.

This is a balance concern because speed is everything in a fight and the downsides are negligible. I've used it to steamroll factions before because I can outrun them completely.

This isn't the actual cola you have in vendors-the REAL nuka cola is intentionally virtually unobtainable, and given that you removed ephedrine speed from actual ephedrine, there's no reason to have a commonly available drink even a raider could make do the same.

### DIFF
--- a/code/modules/reagents/chemistry/reagents/drink_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/drink_reagents.dm
@@ -392,18 +392,6 @@
 	glass_name = "glass of Nuka Cola"
 	glass_desc = "Don't cry, Don't raise your eye, It's only nuclear wasteland."
 
-/datum/reagent/consumable/nuka_cola/on_mob_add(mob/M)
-	..()
-	if(isliving(M))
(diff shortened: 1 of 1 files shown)


UndeadNappist:
It's interesting that stimulum does this and it's both a reagent and a gas.

AdamElTablawy:
Stimulum is impossible to make without fusion and i'm the only person on the server who can do fusion, so it's not an issue.

improvedname:
Adam is right, not forgetting the fact you can't make canisters, it be nearly impossible to manage.
Stimulum is also a reagent because it gives you a special effect stun immunity

rhicora:
This looks okay and I don't see strong objections.

PR #585, opened 2019-02-16
Nerfs several alcohols so that they are no longer incredibly lethal (bye neurotoxin)

AdamElTablawy:
Chem combat is detrimental to FO13 as a whole but who cares for the most part because let people shoot each other with stupid shit instead of ballistics if they want, whatever.

The issue is that neurotoxin and beepsky smash are both instant stuns that last as long as the chem is in your body, and, accordingly, are the reason that chem deathmixes are so lethal-the person cannot respond in any way or save themselves, and you can kill them at your lesiure-or the mix will do it for you.

Given that FO13 is virtually devoid of instant stuns as compared to regular SS13, this makes neuro a serious problem. Shotgun darts have proved that several times-it's not very difficult to immediately make neurotoxin, load it into a shotgun dart, and blast someone with neuro + any pyrotechnic chem.

I'll see if pyrotechnic chems still need a nerf after this, but honestly their main issue and the issue with chem fighting in general is just the stunlock that comes from this shit instantly being complete BS and leaving you unable to fight back at all.

Otherwise, brave bull giving you an extra +10 health pairs too well with the lifegiver perk-that's an extra +25 health total and that's sort of insane.

Hearty Punch is basically an atropine that can be made in the bar and combined with atropine, and given the previous chemistry nerfs, shouldn't be as a result.

Other bar drinks are untouched-i'm fine with people healing from alcohols. Just not the insane level that Hearty Punch obliges.

### DIFF
--- a/code/modules/reagents/chemistry/reagents/alcohol_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/alcohol_reagents.dm
@@ -545,18 +545,6 @@ All effects don't start immediately, but rather get worse over time; the rate is
 	glass_icon_state = "bravebullglass"
 	glass_name = "Brave Bull"
 	glass_desc = "Tequila and Coffee liqueur, brought together in a mouthwatering mixture. Drink up."
-	var/tough_text
-
-/datum/reagent/consumable/ethanol/brave_bull/on_mob_add(mob/living/M)
-	tough_text = pick("brawny", "tenacious", "tough", "hardy", "sturdy") //Tuff stuff
-	to_chat(M, "<span class='notice'>You feel [tough_text]!</span>")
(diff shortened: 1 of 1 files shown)


improvedname:
I support

AdamElTablawy:
thanks

RenGusta:
Rational.

rhicora:
This seems fine to me but I'll keep it up for a day before merging in case there are any objections.

UndeadNappist:
It's kind of interesting that neurotoxin has all these effects under the datum, along for various thresholds. Evaluate or switches would likely work better in the future.

rhicora:
This looks okay and I don't see strong objections.

PR #587, opened 2019-02-16
rnd wip stuff

improvedname:
will put all details when i finish

-Changelog-
Tweak:Renames Bluespace RPED, to Deluxe RPED which doesnt have the ability to upgrade from distance but just a superior carrying capicity.
Add:Adds a base for loot design disks

@GhostEcho 
If you could do give this a test when this is done, it will be appreciated.


UndeadNappist:
This is one of those things that just makes QoL better for those ~~dirty powergamers~~ people who prefer research but doesn't necessarily break the game. Gonna need an explanation as to why.

ErwinSnowden:
Add Ultra Bins as Bluespace bins if you gonna do that chief

improvedname:
Gonna hook this up to the new signals c o d e

PR #589, opened 2019-02-17
Removes magicarps

improvedname:
So you cant turn the wasteland into this monsteriousity anymore
image

### DIFF
--- a/code/modules/events/wizard/magicarp.dm
+++ b/code/modules/events/wizard/magicarp.dm
@@ -36,6 +36,7 @@
 	projectilesound = 'sound/weapons/emitter.ogg'
 	maxHealth = 50
 	health = 50
+	gold_core_spawnable = FALSE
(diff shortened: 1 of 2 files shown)


AdamElTablawy:
can still create gutsies with lp slime potions atm

would be safer to remove gold reactions entirely

improvedname:
@UndeadNappist changed to magicarp only removal requested by teatree

ErwinSnowden:
This was a problem with people griefing and leaving the vault instead of trying to save it, he had yellow slime extracts which could've been used to power the vault. Instead the admins let the server get worse and worse instead of taking action where it was needed, the only reason why the wasteland turned out to become that was because someone was assertive and tested their chances, gee it's almost like the admins wanted it to happen so they could remove more xenobio content instead of tweaking it. :thinking:

PR #593, opened 2019-02-17
Update simple_animal.dm

UndeadNappist:
Maximum cows cannot be exceeded.
Mind giving this a look over? @rhicora

### DIFF
--- a/code/modules/mob/living/simple_animal/simple_animal.dm
+++ b/code/modules/mob/living/simple_animal/simple_animal.dm
@@ -108,7 +108,7 @@
 	NPC_count += 1
 	if(NPC_count > MAX_NPCs)
 		qdel(src)
-		visible_message("<span class='danger'>Die, monster! You don't belong in this world!</span>")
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/simple_animal/simple_animal.dm
```diff
@@ -101,6 +105,10 @@
 		real_name = name
 	if(!loc)
 		stack_trace("Simple animal being instantiated in nullspace")
+	NPC_count += 1
+	if(NPC_count > MAX_NPCs)
+		qdel(src)
+		visible_message("<span class='danger'>Die, monster! You don't belong in this world!</span>")
```
rhicora: Does this need to be a visible message to everyone? It's kind of an OOC deletion.

rhicora:
A global NPC cap is a very good change but I would make the message an admin log instead of a public-visible message.

code/modules/mob/living/simple_animal/simple_animal.dm
UndeadNappist: Only admins get meme humor now.

UndeadNappist:
Only admins are able to see the meme now.

PR #598, opened 2019-02-18
Update simple_animal.dm

UndeadNappist:
It's literally one variable guys, yikes. #597

### DIFF
--- a/code/modules/mob/living/simple_animal/simple_animal.dm
+++ b/code/modules/mob/living/simple_animal/simple_animal.dm
@@ -92,7 +92,7 @@
 	var/my_z // I don't want to confuse this with client registered_z
 
 	//Stops the game from crashing
-	var/const/MAX_NPCs = 255
+	var/const/MAX_NPCs = 2500
 	var/global/NPC_count = 0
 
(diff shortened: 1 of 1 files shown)


Autismobirb:
If the number is raised that high there is no point having that section of code, servers wasting time counting chickens.

For the time being the simple request was to remove this section for the time being, much like Rhi had done in #597

UndeadNappist:
One extra NPC is more for the server to keep track of than a single counter.

rhicora:
This looks fine from a code standpoint.

improvedname:
But can you edit the var?

rhicora:
@Autismobirb I think there's a fair case for having the mob limit as a sanity test; if someone is spamming cows / slimes to the point where it starts to drag on the server, we should stop it at some stage.

UndeadNappist:
If the limit gets hit again and is irritating, it can be raised again. This also coupled some code cleanup with it.
My major concern is that someone __will__ - And I don't mean might, the server is up for 24/7 - eventually cause a slimepocalypse or something else equally low tier and ridiculous, given that they scale geometrically.

PR #599, opened 2019-02-18
Restart vote passing, with admins on now autocalls the shuttle

improvedname:
IF THE VOTE PASSES WITH ADMINS AND CODERS ON IT AUTOCALLS THE SHUTTLE
RESTART VOTE WITH NO ADMINS/CODERS ON IS STILL INSTANT FOR OBVIOUS REASONS.


rhicora:
Restart should probably still be a shuttle call for when admins are not on.

improvedname:
Normal restart isn't covered by this, for severe issues like lag when theres no admins/coders on

rhicora:
That's a good case for fast restart when no one's on. I'll approve this.

PR #610, opened 2019-02-19
[Balance] Quirks

Raze500:
## Description
The changes are:

- Deviant tastes cost now +1
- Family Heirloom is now free!
- Social anxiety now cost -2 [This quik is so shit that people not use it]
- Reduces  the lifegiver bonus to +10
- Ramdom bodypart now cost +1 beacuse you can heal it by just using cables and welder! 

`This maybe be all it's 10:50 right now and im going to sleep, if i fogort something, check the code muh.`

## Motivation and Context
Anti-Powergeimeer measure!

## How Has This Been Tested?
Tested and Hosted

### DIFF
--- a/code/datums/traits/good.dm
+++ b/code/datums/traits/good.dm
@@ -80,7 +80,7 @@
 
 /datum/quirk/lifegiver/on_spawn()
 	var/mob/living/carbon/human/mob_tar = quirk_holder
-	mob_tar.maxHealth += 15
+	mob_tar.maxHealth += 10
 	mob_tar.health += 15
 
 /datum/quirk/light_step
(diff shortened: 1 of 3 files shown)


JTGSZ:
All I have to ask is, what was going through your head when you made this one?

Raze500:
People always try to get family heirloom to get the lifegiver quirk.

Deviant tastes gives a big good moood eating organs.

People should get social anxiety bcuz why not

Also everybody knows family heirloom is a free bonus

JTGSZ:
You do know you need eye protection to not burn your eyes out while working on a prosthetic limb, which makes it a pain in the ass to deal with since you will carry that around if you plan on healing it.

Why would anyone even bother taking family heirloom now? Its actually annoying to deal with especially since you drop and forget the item sometimes.

Deviant Tastes shouldn't require a point to buy. 

Lifegiver is fine being nerfed, or you know boosted up in points.

People already get social anxiety, personally in my opinion I don't like this set of changes as its just value changes on whats already there. It also promotes people not bothering with what used to just be a niche system on character creation at all.

But, someone else can review and vouch for this.

kilozombie:
Almost entirely awful changes, let's look through 'em

>Deviant tastes cost now +1

Deviant tastes makes certain things like human organs tasty, but food like fruit and chocolate untasty. This isn't really a benefit or disadvantage, that's why it's a neutral trait.

>Family Heirloom is now free!

While I understand what you're trying to do here, this is an ugly, bad fix. Now you can take Family Heirloom but you'll still get the mood disadvantage for losing the object?

>Social anxiety now cost -2 [This quik is so shit that people not use it]

You say this is to reduce powergaming, but taking a quirk that only affects speech so you can get Lifegiver sounds like a great powergaming option. Very dumb.

>Reduces  the lifegiver bonus to +10

Okay.

>Ramdom bodypart now cost +1 beacuse you can heal it by just using cables and welder!

No?

As JTGSZ said, repairing one of these is nontrivial. In addition, they also have less health, and as far as I'm aware they also pop off the moment you get EMP'd. I'm pretty sure you walk slower with a prosthetic leg, too.

There's already plans for a larger rework of the quirk system, so this PR really doesn't seem productive, well-thought-out or even remotely necessary.

Raze500:
Family heirloom is annoying yes, but whatnot -1 is good to get lifegiver or big leagues... and also you cant be so dumb to lost your heirloom... just make a box and done isnt hard.

Deviantastes it should, it makes you not making you puke for eating raw meat *its like a goddam cannibal trait*

ErwinSnowden:
add technophreak for +4 points and im sold

UndeadNappist:
#488 
Moods being bad need to be made worse. Append a slowdown effect to it.

Hallucinations already occur at low numbers.

UndeadNappist:
I think fundamentally the issue is that this encourages min-maxing and powergaming. I don't want that. Even if it may be, mechanically, more balanced.

I want people to pick traits because they want to play those traits.

Not because it gives them an advantage, ala big leagues and lifegiver;
And there has been a past discussion on changing the quirks system to a perks system like the Fallout games use, where each perk comes with it's own disadvantage (one-handed, for example).

PR #611, opened 2019-02-19
nerfs arm-mounted implant toolsets

AdamElTablawy:
no point in printing advanced or alien tools if you can just hotglue it into your fucking arm lmfao

seems like an unnecessary powergamer buff just like the advanced machine builder was, and if alien tools in general are still being a problem i'll probably remove those research nodes entirely but as it stands augument toolsets should be T2, not T4, tools.

### DIFF
--- a/code/modules/surgery/organs/augments_arms.dm
+++ b/code/modules/surgery/organs/augments_arms.dm
@@ -177,8 +177,8 @@
 /obj/item/organ/cyberimp/arm/toolset
 	name = "integrated toolset implant"
 	desc = "A stripped-down version of the engineering cyborg toolset, designed to be installed on subject's arm. Contains all necessary tools."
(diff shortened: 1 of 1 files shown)


ErwinSnowden:
I'd prefer this option of edging towards removing advanced tools in general so we can eventually get balanced abductors in, or just give abductors an omnitool which instantly does all the functions of the regular tools but that'd be fucking stupid as we already have content for abductors, it'd be pretty powercreepy.

GhostEcho:
Tbh I prefer them being T1 tools for sacrificing effeciency at the price of not having to lug tools in ya belt or bag

AdamElTablawy:
@GhostEcho they are t2 on tg and they will stay t2 and if you disagree fuck you

AdamElTablawy:
@ErwinSnowden no

improvedname:
@AdamElTablawy can you keep surgical tools for the arm, unless you port tg's surgical powertools

improvedname:
Will be adressed in my rework instead

PR #612, opened 2019-02-19
Adds gun slowdown

UndeadNappist:
**Description of changes:**
- Adds Paralyze
- Implements a very short period of paralyze when you draw a gun based on the weapon's weight
- Adds a slowdown when you have a gun OUT and WIELDED and only then.
- Adds to your attack log whenever you grab a gun.
- Other planned but not yet implemented features.
- Similar plan for melee weapons.

**Checklist before we merge your thing:**
* [x]  The author is ready to merge
* [x]  All changes are described above
* [x]  24 hours for comment have passed, or this is a bugfix
* [x]  Compiled
* [x] Hosted and Tested.


code/modules/projectiles/gun.dm
```diff
@@ -126,7 +125,17 @@
 				user.visible_message("<span class='danger'>[user] fires [src]!</span>", null, null, COMBAT_MESSAGE_RANGE)
 
 
+//Adds logging to the attack log whenever anyone draws a gun, adds a pause after drawing a gun before you can do anything based on it's size
+/obj/item/gun/pickup(mob/living/user)
+	. = ..()
+	user.visible_message("<span class='danger'>[user] grabs a gun!</span>") // probably could code in differences as to where you're picking it up from and so forth. later.
+	user.Paralyze(2 * src.weapon_weight, ignore_canparalyze = TRUE) // Need to define where you're grabbing it from, assign numbers to them, and then divide the paralyze total by that. Tables/holster/belt/back/container.
+	user.log_message("[user] drew a gun", INDIVIDUAL_ATTACK_LOG)
+	src.slowdown = (w_class / 2)
```
UndeadNappist: note to self, bind this all into a proc under obj/item
weapon_weight might be a problem though

UndeadNappist: fuck you past me I thought I was done

UndeadNappist:
I recommend merging this right away. While VERY TECHNICALLY I still need to do things - reinsert the /weapon/ parent type, do a mass replace+change, add some other details, tweak this with melee weapons ; the QoL affect from having this in cannot be understated.

Also, I want feedback on how this feels in active, chaotic combats.

GhostEcho:
Why a paralyze? If it uses the channeling it should cancel when you move, this makes it so you're locked into drawing the gun out

UndeadNappist:
Exactly. I want it to be a commitment.

improvedname:
Only make it log when you auctally draw it, just use channeling

UndeadNappist:
> Only make it log when you actually draw it, just use channeling

Then we will have problems with people starting to draw a gun, getting shot, then complaining in adminhelps the other person didn't properly escalate and it'll boil down to a he-said-she-said argument, as I am already trying to solve with this PR.

improvedname:
Eh its fine to expiriment with your right

PR #332, opened 2019-01-02
Remove ERP verbs

rhicora:
Rationale for this change:

- The use of the verbs encourages LRP gameplay
- The descriptive writing surrounding the verbs is low quality and repetitive
- The use of the verbs regularly replaces high-quality emotes
- The verbs have no notion of OOC consent, which routinely violates Rule 1
- The verbs were introduced solo by the previous host and did not receive code review

PR #618, opened 2019-02-20
Improve lewdcode quality

rhicora:
If you can't beat 'em, join 'em. Substitute for removing ERP verbs (#332), makes them better and more varied in their description. 

The breast grope action offers an example of intent-based ERP, we can expand this to other verbs if need be. Other changes:

- Divorces sex organs from gender. 
- is_topless() and is_bottomless() are now separate concepts.
- Adds dickbiting / retaliatory damage against partner in Harm mode.
- Adds castration so you can remove Bear testicles.
- Applies gender_ambiguous only when breasts mismatch with gender presentation, or surgery was messed up.


UndeadNappist:
+FUN

You sometimes can't fix behavior, but you CAN punish them for it.

Raze500:
Hmm yes i can finally spam click SEX.

PR #631, opened 2019-02-21
Make ghouls hostile to ghoul players

improvedname:
sorry bud its kinda unbelanced, when you kinda can bypass 70% of the areas without problem

Unique Races should be for RP flavour with slight mechanical difference, but this is just too severe

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/ghoul.dm
+++ b/code/modules/mob/living/carbon/human/species_types/ghoul.dm
@@ -20,10 +20,10 @@
 	..()
 	for(var/obj/item/bodypart/b in C.bodyparts)
 		b.max_damage -= 10
-	C.faction |= "ghoul"
+	C.faction |= "pghoul"
 /datum/species/ghoul/on_species_loss(mob/living/carbon/C)
 	..()
-	C.faction -= "ghoul"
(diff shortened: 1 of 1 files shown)


Raze500:
Raze did an oopsie

GhostEcho:
Rhicora a bit speedy of a merge with no discussion on it besides Raze aaaaa

GhostEcho:
Discussing with Jt in general a bit on this, it's not as severe as it seems at first glance, as the only area where this is *really* an issue is the prison due to the good weapons and armor that can spawn there, however that area is easily dealt with already with even just a pistol, so ultimately being able to bypass ghouls isnt an issue. A better solution would be to replace the ghouls in the armory of the prison with some protectrons which will attack ghoul chars and are ultimately a better defense for a loot stockpile on par with the east enclave bunker

JTGSZ:
Pretty much, I thought ghoul species having the unique advantage of being able to manipulate ferals was actually you know, interesting.

Not everything has to be a reskinned boring human, there can be differences. Not to mention the bright brotherhood used their feral friends to fight nightkin.

But, I feel like it should return in a PR in the future for now, as I have a few actual good plans for them.

PR #642, opened 2019-02-22
Ghoul Hair and makes feral ghouls friendly to ghouls again.

Raze500:
- [x] Ghouls can now have hair

- [x] Player ghouls are friendly now with feral ghouls.

Knowing that #644  will work and tweak most of dungeons of adding servitrons/bots/other bad robots, i think is safe to merge this that will add a new mechanic to the people that play ghoul of course.

And if you say ghouls are OP for making them friendly to ferals, try to kill a ghoul player with a *normal* fucking saw and JAJA's in spanish i guess.


UndeadNappist:
You made them faster as well. Please don't try to slip rebalancing into graphical updates without stating them and without discussion.
#644

PR #643, opened 2019-02-22
Animal and Raider Migration - Also APC Area changes

GhostEcho:
Some scorpions have appear to migrated towards areas, attracted to the scent of people frequenting these areas. Raiders have also appeared to take over the super mart from it's previous ghoulish inhabitants, bringing in some meds, drugs, and average weaponry. Someone has also installed some more robust security in the prison armory, or has it always been this way?

Also prepares apc areas to be more sensibly constrained to buildings, and for sand storms looming on the horizon


GhostEcho:
ah fuck me
map conflict

PR #645, opened 2019-02-23
Adds Throwing spear recipe and Head on pike recipe with Throwing spears.

GoodG1448:
Legion should be able to do something with the heads of their enemy when they get out of hands with their machetes. Again, not sure if I did it right, might need testing.

### DIFF
--- a/code/modules/crafting/recipes.dm
+++ b/code/modules/crafting/recipes.dm
@@ -567,6 +567,16 @@
 	result = /obj/structure/headpike/bone
 	category = CAT_PRIMAL
 
+/datum/crafting_recipe/headpikelegion
+	name = "Spike Head (Throwing Spear)"
+	time = 65
+	reqs = list(/obj/item/throwing_star/spear = 1,
+				/obj/item/bodypart/head = 1)
+	parts = list(/obj/item/bodypart/head = 1,
+			/obj/item/throwing_star/spear = 1)
+	result = /obj/structure/headpike
+	category = CAT_PRIMAL
+
 /datum/crafting_recipe/smallcarton
 	name = "Small Carton"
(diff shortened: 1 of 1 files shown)


UndeadNappist:
Theoretically you should be able to just go to the original headpike recipe and make it list( /obj/item/twohanded/spear = 1 || /obj/item/throwing_star/spear = 1 || /obj/item/twohanded/bonespear = 1, /obj/item/bodypart/head = 1)

At least, I hope so. It would also cut down on lines of code.

Add a recipe for legion spears too. Make it require sheets of something special.

GoodG1448:
I didn't wanted to overwrite the glass spear one since that would make it ankward for vaulties to make heads on pikes with available materials inside the vault. For displaying the Overseer head in front of the office and related shenanigans.

About the legion spear recipe, should I use iron ore instead of metal sheets?. I was thinking about something like 2 metal sheets and 3 rods or 1 metal ore and 3 rods (Both seem kinda cheap imo, it would be rarer with the ore one since it is not that common to find).

UndeadNappist:
> I didn't wanted to overwrite the glass spear one since that would make it ankward for vaulties to make heads on pikes with available materials inside the vault.
|| is an or statement. **_In theory,_** you could just slap the or statement down and the crafting recipe would check for either object and allow you to craft for it.
It would save time, space, and processing power, along with tidier code and less spaghetti.

> EDIT: Now that I think about it, vaulties would have access to throwing spears since they can get truckloads of iron ore with no sweat. Maybe adding a leather sheet?
Sure, sounds great. Sinew maybe, since it's in.

GoodG1448:
Ready for merge. I guess.

PR #649, opened 2019-02-23
Power station made useful

KanohaShinobi:
## Description
This adds solars, solar controller, and a new room in the power plant. In this room is the APC's controlling the NCR base, the power plant, sunnydale, then den, farms, and outer Klamat (not the mine). 
## Motivation and Context
This Makes the Power Station actually useful and not just a big building of spare generators full of ghouls. IS also a balance thing, since puts power of a few places into anyone's hands (likely the NCR's since they're right next to it). But it gives the NCR and the Legion something to fight over, as well as the NCR some grip on the den without trying to annex the shit hole, cuz thats a lost cause. An Over all objective point that can help make some rounds interesting.

## How Has This Been Tested?
Hosted the thing (invisible of course) and tested it. Had to fight through ghouls but got to the room regardless. The solar thing needs to be refreshed at the start but SMES' power everything for a bit at round start. The power flow control is permanently locked since no one in the wasteland has the ID for it, and can't get it (I dont think). If they did have an ID though, it controls Vault powers as well so I didn't leave a way of access.

##NOTE:
 This is supposed to be merged after ghost's map update "Mob Migration 2 Electric Update Edition" (since uses the same map to prevent conflicts). Map was only edited thing.


companoo:
probably a better solution then having all the power in a mountain

KanohaShinobi:
It doesn't make sure it's in the hands of the NCR. The den's people could easily take the power station (Or even take it from the NCR) as well as the legion. I was hoping it would give the den something to do to protect itself rather than just sit on it's ass with the gate closed. But I guess will make it not control den.

GhostEcho:
Now Nappist thats a rather incorrect analysis, as you're making the false assumption that A: the Den is a prize, and B: that the NCR cant just do what you suggest already

KanohaShinobi:
The "power" in the hands of the NCR makes sense. Much Like it being in the hands of the NCR in the mojave. They control both Helios One and the Dam. NCR sees power they rush to it. Mainly cuz they can send it back home, or tax people for it. Also tbh as NCR it's always wisest to leave the den alone because its usually just a bar with a floor soaked in blood. Same with the streets. Den is mostly a lost cause for the NCR. And there probably would have to be  rule that the person controlling the power couldn't just leave it off for the den, since the den needs it for the blast doors. So I'll remove den from the power plant to avoid causing further retardation.

PR #651, opened 2019-02-23
Adds some fallout posters

KanohaShinobi:
## Description
adds some actual fallout posters from fallout New Vegas (all NCR except 1 which is the corporate espionage one).

## Motivation and Context
Solves a few of the gay ass posters you see on walls that are all still tg station and not lore friendly. Gives some propaganda for the NCR to have in their base (and i later might add posters that are the object that the NCR can go hang up on the map)

## How Has This Been Tested?
Tested when hosted and spawned them.

##Screenshots
http://prntscr.com/mp81ka


UndeadNappist:
Does this have a dependency on #649 or the other way around, or are they irrelevant to each other?

KanohaShinobi:
No, its not a map change or anything. It just adds the object. So posters won't be in the map at all, YET. For now i guess admins etc. would be able to spawn them in, but thats it. Just adds some posters and shouldn't conflict with any other PRs unless they edit contraband.dmi and .dm

PR #655, opened 2019-02-25
Maintenance[Someone actually check this]

JTGSZ:
## Description
A giant mass of upstream merges and ports that consist of mostly optimization, bugfixes, refactoring, reworks.
Delay to voting is added, variable can be changed in subsystem vote.dm, this stops lobby votes.
Its currently set to 30 minutes.
Lights now flicker when you hit and don't break them.
This stops at July 18th/19th as BSQL needs implemented before anyone should continue upstream.

## Motivation and Context
BD needs maintained, there were a lot of resolved issues.
I don't think anyone enjoys dealing with trying to solve all the runtimes anyways.

## How Has This Been Tested?
Hosted to check for initialization runtimes regularly, tested thoroughly on occasion on dodgy looking things.


UndeadNappist:
https://github.com/tgstation/tgstation/pull/38323

Raze500:
This will improve the lag?

JTGSZ:
Maybe.

GhostEcho:
Well even without being fully setup to take advantage of all of it, not a bad idea to merge and begin taking advantage of it unless it *would* break things

JTGSZ:
It shouldn't.

PR #656, opened 2019-02-25
adds ncr ranks, changes title of captain and lieutenant, and adds an NCR quartermaster

companoo:
does what it says on the tin

adds an NCR qm role with the rank of ssgt

goes like this

CO = CAPT
XO = LT
QMSGT = SSGT
SGT = SGT etc etc

https://imgur.com/a/vUtSwct screenies

i tested everything i think it all works

check my code please


RenGusta:
Did you make sure it doesn't fuck up timelocks and they're properly configured for it, and don't invalidate experience on the roles they're replacing?

I also don't know why you'd even need to bother changing Captain and Lieutenant to CO and XO respectively. It's easier to understand that there's a Commanding Officer and his Second in Command, than renaming the roles something less intuitive.

companoo:
timelocks should remain working, have been configured so that QM takes a bit more then sergeant, check fail is because i was using an outdated map to set up the spawn, can fix that easy, CO and XO can be changed but i think it looks better

companoo:
ill shut this and fix map real quick

PR #658, opened 2019-02-26
adds ncr rank pins, ncr QM

companoo:
basically #656 

-cap and lieutenant stay as they are
-qm sergeant is now just qm
-ranks are there so that in future we might be able to do some cool stuff with them

screenies: https://imgur.com/a/vUtSwct

review this shit im new


UndeadNappist:
Use the BoS dogtags as a template.

improvedname:
No wait, we have normal tags in the code, use that instead

companoo:
this doesent change ID's it just adds an accessory onto every NCR guy showing there rank when examined, and the NCR QM role, not gonna touch ID's

JTGSZ:
Yeah, accessories like pocket protectors/skull codpieces/medals. These can be clipped onto uniforms which is hella cool.

PR #659, opened 2019-02-26
Adds coal -Might need unfucking-

GoodG1448:
I guess this will do for now. Change the values as you see fit. Also, should we deleted the volcanic and ice rockgens for less clutter?

Could also change the grind result of the coal lumps from carbon to charcoal to give it some use as an alternative medicine for toxins.


code/game/objects/items/trash.dm
```diff
@@ -76,9 +76,11 @@
 	icon = 'icons/obj/mining.dmi'
 	icon_state = "slag"
 	desc = "Someone's gotten on the naughty list."
-	grind_results = list("carbon" = 20)
+	grind_results = list("charcoal" = 20)
```
UndeadNappist: Charcoal is made from wood. Coal is made from carbon. They are two different substances.

code/game/turfs/simulated/minerals.dm
```diff
@@ -218,6 +218,12 @@
 
 
 
+/turf/closed/mineral/coal
+	mineralType = /obj/item/trash/coal
```
UndeadNappist: Do NOT use /obj/item/trash/coal as your mineralType ; it will break a lot of things. Use mineral sheets. Yes, this will require making a new object.

GoodG1448:
Well, the /obj/item/stack/ore file is impossible to find so I can't really add a new ore to it. About adding charcoal to it, not sure what is the point of the coal addition, I was thinking that in the future it would have some uses for making gunpowder (By mining coal, saltpetre and sulfur) so I changed the regeant to charcoal.

UndeadNappist:
> Well, the /obj/item/stack/ore file is impossible to find

It's not a file. You will have to make a new object type and tool around with things to get it to work.
https://github.com/BadDeathclaw/TG-Claw/blob/7d5396e109f0606fe90ad3d6c2660784c7fe5613/code/game/objects/items/stacks/sheets/mineral.dm

>I changed the reagent to charcoal. (Which makes black powder when mixed with salpeter and sulfur, then heated)

Once again, coal is made from carbon. Charcoal is made from burning wood. https://en.wikipedia.org/wiki/Charcoal

GoodG1448:
Try to figure it out yourself

GoodG1448:
Hey this may sound retarded, but what If we actually give the coal ore an actual use and turn it into plastic once smelted in a furnace/ORM?. Since plasma is kinda non-existant in fallout lore, what about replacing the plasma ore with the coal ore. Or instead of just replacing it, make plasma a lot rarer than coal.

If this would be the case, then a lot of the protolathe recipes that required plasma would have to be changed to plastic.

GoodG1448:
Well, after days of taking a look to the whole code. This is over me, no clue on how to make a new ore.

If anyone wants to continue this, go ahead.

PR #665, opened 2019-03-02
Raider loot and some balance changes

GhostEcho:
Well figured out how to make raiders drop weapons, as part of the raider boss change, he now drops a 9mm smg, the ranged ones drop mosins, and melee ones drop sledgehammers

As such, some changes to them have been changed 
Melee boys have 140 health instead of 120 and do between 30-56 damage, as they're using sledgehammers
Regular range boys have 115 health and shoot appropriate ammo type, so 7.62 rounds
Raider Boss is up to 170 health


Raze500:
IT JUST WORKS

PR #670, opened 2019-03-04
misc

improvedname:

### DIFF
--- a/code/game/objects/items/inducer.dm
+++ b/code/game/objects/items/inducer.dm
@@ -9,7 +9,7 @@
 	force = 7
 	var/powertransfer = 1000
 	var/opened = FALSE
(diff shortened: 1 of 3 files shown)


ChronicPwnage:
The sugar change seems strange and unneeded, it basically turns 30u of sugar into an instant knockout pill, could abuse sugar smoke, sugary foods, poisoning etc in many many ways. Why change it to 20u to overdose?

ErwinSnowden:
just shoot somebody with sugar syringe gun and date rape them now pill cosby... this is a bad change and negatively effects balance because I can get the lethal injection syringes, empty them, fill em with sugar and in the syringe gun it goes. Now I've just made a ranged instant KO that lasts ages. Now I can rape whoever doesn't have thick armor on.

Raze500:
6 days ago

PR #675, opened 2019-03-09
Revert "Blood cascade fix"

companoo:
it breaks everything

ren says so


UndeadNappist:
Unless the server was updated with the PR - which I doubt - or it was tested on a private server with these changes - the statement about it breaking everything is inaccurate.

The changes have not hit live regardless.

PR #685, opened 2019-03-12
Fixes softcrit whispering

UndeadNappist:
@JTGSZ you no longer die while whispering in soft crit, only in hard crit.

The whole thing could probably be refactored to be under MESSAGE_MODE and the like, but I ain't doing it.


JTGSZ:
I tested this, I was dying in any crit. By Softcrit you do mean when you are able to crawl around on the ground. Yeah?

UndeadNappist:
Should be, yeah.

GhostEcho:
So now softcrit lets you talk over the radio and doesn't register as "Barely conscious"

UndeadNappist:
Fuck.

Issue #640, opened 2019-02-22
Add craftable bows


Raze500:
- [x] Crafteable bows

- [x] Crafteable crossbows

- [x] Crafteable arrows

- [ ] Adds bows has a ramdom loot(?)

- [ ] Make raze trialcode in the serve lul

Kimochiiiiiiiiiiiiii:
+1 to everything but the last one.

improvedname:
+2 to the last one

Raze500:
The bow code is ready i just need the sprites and know where put them

And also i need make the arrows code.

Insanegame27:
I can try and do some spriting for ya, but I need to know more details regarding theme, mechanics, materials etc. compound, recurve, self? Are there multiple types? What poundage draw weight is being abstractly emulated?

Raze500:
For now i guess would be great a normal bow, from hood and metal... the arrows and a quearry to have the bows.

Some sprites in hand right/left and the proyectile too would be nice.

Kimochiiiiiiiiiiiiii:
There's crossbows already, a map spawns with a vox pirate base that has it

Raze500:
B o w s

Kimochiiiiiiiiiiiiii:
You said crossbows too, fuck off.

Raze500:
I just need some #Nerd Codder monkey will add it has ramdom spawn and some coder mother make me trial coder *SCREAMS IN SPANISH*

PR #695, opened 2019-03-14
Tribal Update: Crafteable Arrows/Bows and quivers!

Raze500:
Co-Authored-By: imw-Oculus <imw-oculus@users.noreply.github.com>
CHECK #640 
## Description
Adds crafteable bows/arrows and quivers.

## Motivation and Context
I like bows.

## How Has This Been Tested?
Yes, It works good and even the crafting! 
Of course not ghouls died on this testing!!!11!
And also somehow the gunner bandit are inmune (A bug i guess)

## Something else to add?
The quiver needs new sprites and the arrow too, sprites please come here and aid this spanishtard :^)

## What is new?
-The arrows deals 30 brute DMG (Not in hand, just when you shoot it!) *in hand it deals like 5-10 dmg i think*
-The quiver hold **20 A R R O W S !**
-Adds the bow and it gives a thicc sprite with the arrow too! (the quiver needs the resprite)

image


JohnOxford:
quiver

better quiver
arrow

better arrow

Raze500:
Thanks!!!!!

code/modules/projectiles/guns/ballistic/bow.dm
```diff
@@ -0,0 +1,102 @@
+/obj/item/gun/ballistic/bow
+	name = "bow"
+	desc = "A sturdy bow made out of wood and reinforced with iron."
+	icon_state = "bow_unloaded"
+	item_state = "bow"
+	var/icon_state_loaded = "bow_loaded"
+	var/icon_state_firing = "bow_firing"
+	var/item_state_loaded = "bow"
+	var/item_state_firing = "bow"
```
UndeadNappist: Why are these the same? Why even have either one if the item_state never changes?

code/modules/projectiles/guns/ballistic/bow.dm
```diff
@@ -0,0 +1,102 @@
...
+	fire_sound = "bowshot" //
```
UndeadNappist: .ogg

code/modules/projectiles/guns/ballistic/bow.dm
```diff
@@ -0,0 +1,102 @@
...
+	mag_type = /obj/item/ammo_box/magazine/internal/bow
+	//var/flags = HANDSLOW //
+	weapon_weight = WEAPON_HEAVY  //
+	var/draw_sound = 'sound/weapons/draw_bow.ogg'
+	var/ready_to_fire = 0
+	var/slowdown_when_ready = 2
```
UndeadNappist: it's already got an implemented slowdown, this makes it worse. I think it'll make it feel really bad.

code/modules/projectiles/guns/ballistic/bow.dm
```diff
@@ -0,0 +1,102 @@
...
+		slowdown = slowdown_when_ready
+	else if(magazine.ammo_count() && !ready_to_fire)
+		icon_state = icon_state_loaded
+		item_state = item_state_loaded
+		slowdown = initial(slowdown)
+	else
+		icon_state = initial(icon_state)
+		item_state = initial(item_state)
+		slowdown = initial(slowdown)
+
```
UndeadNappist: Could probably make a switch of some kind and instead of force-setting the slowdown, set it to be slowdown = slowdown + slowdown_when_ready.
for example
firing_status = 0, 1, 2, etc and then define the various states and work from there.

Careful when altering procs that are very high up the parent.

code/modules/projectiles/guns/ballistic/bow.dm
```diff
@@ -0,0 +1,102 @@
...
+/obj/item/gun/ballistic/bow/attack_self(mob/living/user)
+	if(!ready_to_fire && magazine.ammo_count())
+		ready_to_fire = TRUE
+		playsound(user, draw_sound, 100, 1)
+		update_icon()
+	else
+		ready_to_fire = FALSE
+		update_icon()
+
```
UndeadNappist: could probably bind all this into a single /proc/bowshot instead of repeating it for all attacks. I mean, you can suicide with guns.

Unless you meant to make this attack_hand? Or some sort of other item based update.

code/modules/projectiles/guns/ballistic/bow.dm
Raze500: ¿Want a gif to see how it works in game?

code/modules/projectiles/guns/ballistic/bow.dm
Raze500: Item state changes,  it does on the hand!  When you use the bow it changrs the sprite to bow_firing and etc,  later i will make a gif how it works!

JohnOxford:
i think its reasonable for bows to not have a firing cooldown/walkspeed decrease. they're already not that great of weapons and they basically have a forced reload speed of a single shot shotgun, you essentially have to reload after every shot anyways

also we should look into making quivers go on the belt slot, no one is going to use these if they have to get rid of their backpack for it, i imagine it'd look pretty cool to have a quiver under a backpack too.

also maybe make the arrows stack for easier reloading

the cool factor that comes along with a bow will get stamped out if we make the bow worse than the mosin

JohnOxford:
heres a crossbow too, i sent the dmi in your DM's

code/modules/projectiles/guns/ballistic/bow.dm
UndeadNappist: var/item_state_loaded = "bow"
	var/item_state_firing = "bow"
it doesn't change.

code/modules/projectiles/guns/ballistic/bow.dm
UndeadNappist: https://github.com/BadDeathclaw/TG-Claw/blob/master/code/modules/projectiles/gun.dm#L77

much like how i fucked up here and should have made it

slowdown = slowdown + (w_class / 3) we need something similar here. I should go back and fix that too.

Raze500:
Tomorrow i will do this.

rhicora:
Anyone want to pick this up? I think Raze is no longer with the server.

JohnOxford:
he is, he stated he'd be returning after his ban but i contacted him before my wifi went down for the week and he said it was fine if i took over his PR.

i'll be making a new one after i clean up/finish/merge the rest of mine. including the new crossbows

GhostEcho:
Closing for what Oxford said

PR #698, opened 2019-03-15
Gives veteran legionary+ a machete gladius

JohnOxford:
Gives veteran legionary+ a machete gladius


JTGSZ:
Anyone have any comments as to why this should not be the case?

PR #703, opened 2019-03-16
Adds mood modifiers related to dead allies

ChronicPwnage:
## Description

- Seeing someone of your faction (NCR, Legion, BoS etc) die is bad for your mood
- Seeing a bunch of bodies of people in your faction is bad for your mood
- Burying your dead mitigates this

## Motivation and Context

Headmin asked me to, it's a cool little addition that makes it worthwhile to bury your dead and not leave their bodies lying around.
## How Has This Been Tested?

Tested it on a privately hosted version of the code. In general it went pretty smoothly, but I did have a few weird things happen (nothing gamebreaking, just not noticing specific bodies. Could be associated with me adding factions manually via var editing rather than spawning with a faction). Would appreciate further testing, but it should be good to go.


improvedname:
Make it a mood increase if its the dead bodies of your enemies

PR #704, opened 2019-03-16
Changes min vote requirement 60, changes waittime between votes to 30, make shuttle autocall on succesfull vote with admins on

improvedname:
## Motivation and Context
Positive vote by players
image
Prevents vote spam, changes it to the minimun requirement before starting a vote(an hour)

## How Has This Been Tested?
Value changes. but calling shuttle is working from previous PR, just not a check. but since a cooldown is 30 between votes and shuttle progress time is around 26 mins so i wouldnt be an issue

### DIFF
--- a/code/controllers/subsystem/vote.dm
+++ b/code/controllers/subsystem/vote.dm
@@ -1,6 +1,6 @@
 SUBSYSTEM_DEF(vote)
 	name = "Vote"
-	wait = 10
+	wait = 30
 
 	flags = SS_KEEP_TIMING|SS_NO_INIT
 
@@ -16,7 +16,7 @@ SUBSYSTEM_DEF(vote)
 	var/list/voting = list()
 	var/list/generated_actions = list()
 	var/vote_sound = 'sound/f13/mysterious_stranger.ogg'
-	var/min_restart_time = 30 MINUTES
+	var/min_restart_time = 60 MINUTES
 
 /datum/controller/subsystem/vote/fire()	//called by master_controller
 	if(mode)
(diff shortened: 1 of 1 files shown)


companoo:
any way you can reduce the time it takes for the shuttle to arrive with this PR? 20 is long

improvedname:
Suggest a time

companoo:
probably 10 minutes cheers

GhostEcho:
Revert back to ten minutes or 15, it doesnt need to be 30 for spam delay

improvedname:
its gonna have the 30 delay whether you like it or not

GhostEcho:
Yeah because doing it by player vote *is* a good idea, just change it back or reduce it, half hour delay is excessively long

improvedname:
Well too bad

GhostEcho:
Ill change it later then because that's my only issue with the PR

GhostEcho:
Ill change it later then because that's my only issue with the PR

improvedname:
How about no.

GhostEcho:
How about not making changes that will only lead to increased frustration and boredom

improvedname:
Your the only one having issues with it.

SpaceVampire:
I also have issues with it

GhostEcho:
If people vote spam then they can get reported to an admin, if a vote is close and shit happens like a faction raid killing off a faction raid, that means they have to wait essentially an hour to play as that faction again due to A: 30 minutes before a vote can be called again, and B: 30 minutes before the shuttle will actually go through it's process

improvedname:
Guess what vote spam aint a rule 😂

GhostEcho:
"Don't be a dick"

SpaceVampire:
I also think the shuttle should start at ETA: 10min not ETA: 20min

GhostEcho:
It's a beautiful catch all rule that can be applied to things not specifically discussed in the rules

GhostEcho:
Okay that's actually a better solution

JohnOxford:
god you people get so fucking sweaty

PR #694, opened 2019-03-14
Fixes headcoder rank having admin flags

improvedname:
Adminstration, and coding should be seperated 🤷‍♂️

PR #706, opened 2019-03-17
Fixes headcoder rank having admin(+EVERYTHING) flags #2

GhostEcho:
#694 

It was valid concerns and closed too early.


improvedname:
Doesn't matter rhi isnt a headcoder anyways

PR #710, opened 2019-03-20
removes shitty spraytan

JohnOxford:
literally just breaks spraytans

as opposed to having to remove the bottles on the map that contain spraytan, we just made it literally not work

i expect people to make bug reports about how spraytan doesnt work

### DIFF
--- a/code/modules/reagents/chemistry/reagents/other_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/other_reagents.dm
@@ -327,70 +327,9 @@
 	id = "spraytan"
 	description = "A substance applied to the skin to darken the skin."
 	color = "#FFC080" // rgb: 255, 196, 128  Bright orange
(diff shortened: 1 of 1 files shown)


rhicora:
This is not the correct way to solve the problem. Instead:

Cut a new branch, delete `/obj/item/reagent_containers/spray/spraytan` and all underlying code, then open the map. It will complain that it doesn't know what a spraytan is, and ask you what you want to replace instances of it with. Choose nothing.

JohnOxford:
that causes a map edit, even something as simple as that wont get merged fast enough to justify doing it that way and it will fuck over everyone elses map edits.

rhicora:
The correct fix requires a map edit. Ask around if anyone else is working on a map change then do it.

JohnOxford:
you bring me great pain and suffering

PR #718, opened 2019-03-21
Remove Recon Ranger Role

imginbtw:
To alleviate the chain of command problem every round usually comes into, minimize the Ranger role to just two.


JTGSZ:
hello, you are trying to merge master into a old branch.

JTGSZ:
Untitled

PR #719, opened 2019-03-25
Does some mapping QoL, buffs raiderboss, adds raiderboss armor/helmet + greasegun as a drop

improvedname:
Just confines most crates into spawnable objects by the admins, removes messing with the map to change said crates(since its basic equipment)

Lets the raiderboss drop the raiderboss suit/helmet(just reskinned combat armor), and also the greasegun since he fights with that so it makes sense he drops it.

merges legion slave closets into one, instead of having a closet for each item.

Add standard legion/ncr uniforms to the wardrobes

debloats code


code/game/objects/structures/crates_lockers/f13 closets.dm
```diff
@@ -16,34 +16,103 @@
...
-	new /obj/item/clothing/under/f13/caravan(src)
-	new /obj/item/clothing/under/f13/caravan(src)
+	for(var/i in 1 to 3)
+		new /obj/item/gun/ballistic/revolver/caravan_shotgun(src)
+	return
 
-/obj/structure/closet/f13/leigon/ssandals
-	name = "closet"
+/obj/structure/closet/cardboard/f13/legion/singleshotgun
+	name = "large cardboard box"
```
UndeadNappist: You don't need to name = "large cardboard box" for each one, the name will auto-populate based on the parent due to being a var set higher up

code/modules/clothing/suits/armor.dm
```diff
@@ -499,6 +499,16 @@
 	armor = list("melee" = 25, "bullet" = 25, "laser" = 16, "energy" = 16, "bomb" = 25, "bio" = 0, "rad" = 0, "fire" = 0, "acid" = 0)
 	strip_delay = 60
 
+/obj/item/clothing/suit/armor/f13/combat/raiderboss
+	name = "combat armor"
```
UndeadNappist: Here too.

code/modules/clothing/head/helmet.dm
```diff
@@ -332,6 +332,20 @@
 	dynamic_hair_suffix = ""
 	dynamic_fhair_suffix = ""
 
+/obj/item/clothing/head/helmet/f13/combat/raiderboss
+	name = "combat helmet"
```
UndeadNappist: Ditto with previous.

JohnOxford:
not a change but i have a PR running that adds raider power armor, and was considering giving the raider boss it.

you're free to use it in your PR if that tickles your fancy

code/game/objects/structures/crates_lockers/f13 closets.dm
improvedname: i'll like it to have it anyways for clearifiaction, like we don't need all the special same armour values for each copy when we can use subtypes. its not neccesary but its handy to do so.

code/modules/clothing/suits/armor.dm
improvedname: Same as above

code/modules/clothing/head/helmet.dm
improvedname: Same as above

improvedname:
@UndeadNappist yeah but only difference this isn't roundstart equipment, i could buff the boss healthwise so it atleast takes a mag or two to take him down but hey 🤷‍♂️

Autismobirb:
> Greasegun is a ridiculously good smg. I took it away from NCR srgts for the same reason. Sure we wanna do that?

Its a good gun yes, but it isn't free, you gotta kill the cunt who's surrounded by a couple other cunts to get the pewpew.

code/game/objects/structures/crates_lockers/f13 closets.dm
UndeadNappist: It leads to horribly unnecessary bloat, and leads to an issue where things are hard to find and sometimes you want all the children to change when you change the parent. Hard disagree on policy, long term convenience, and proper code management.

You can also write hierarchies as
/obj/item
                /parent
 parent procs defined here
                              /child
Etc, although I forget the exact details during my phone posting. I recommend getting away from the nondescriptive and nonuseful parents like f13 and making them all useful and applicable, so we don't have weird shenanigans where people spawn in weird things or map them incorrectly.

On a further note you may want to consider adding your own parents - maybe "loot" or "armory" or something and setting every faction and such to use the same derivatives. It would clean things up in the future and reduce the amount of work you have to put in too.

GhostEcho:
Re open when youre working on it

PR #720, opened 2019-03-27
RnD Weapon Changes

GhostEcho:
Goodbye ballistics, you were boring, overpowered, and no one liked you besides powergamers
Adds: 
Tesla Revolver: A 6 shot energy pistol that does 52 damage on an unarmored person with one full chamber, arcs to nearby targets for reduced damage, be careful,  will arc to you too! Yellow gloves can prevent this from harming 
One Point Gravity Manipulator: Shoots a repulsive bolt that will throw things a short distance away from the impact point, try laying traps with spears and bolas! Slightly inaccurate
Temp Rifle: Makes people cold.

### DIFF
--- a/code/modules/projectiles/guns/energy/special.dm
+++ b/code/modules/projectiles/guns/energy/special.dm
@@ -261,7 +261,7 @@
 	icon_state = "freezegun"
 	desc = "A gun that changes temperatures."
 	ammo_type = list(/obj/item/ammo_casing/energy/temp, /obj/item/ammo_casing/energy/temp/hot)
-	cell_type = "/obj/item/stock_parts/cell/high"
+	cell_type = /obj/item/stock_parts/cell/ammo/mfc
 	pin = null
 
 /obj/item/gun/energy/temperature/security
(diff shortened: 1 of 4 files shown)


improvedname:
Gravity gun + x objects = server suffer

GhostEcho:
Itll be fiiiine, maybe

improvedname:
Well lets see how vault will handle this new content

code/modules/projectiles/guns/energy/special.dm
```diff
@@ -261,7 +261,7 @@
 	icon_state = "freezegun"
 	desc = "A gun that changes temperatures."
 	ammo_type = list(/obj/item/ammo_casing/energy/temp, /obj/item/ammo_casing/energy/temp/hot)
-	cell_type = "/obj/item/stock_parts/cell/high"
+	cell_type = /obj/item/stock_parts/cell/ammo/mfc
```
rhicora: Does cell_type expect and object reference, or a string of an object reference?

improvedname: Just remove the string parent has that as basic celltype

GhostEcho: Just went with Hyenas, still works fine, suppose there's some cleanup opprotunities here tbh

PR #723, opened 2019-03-27
Adds changelog to pull request template

improvedname:
You'll see this makes the ingame changelog work so theres no longer need for the dev-announcements

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -17,3 +17,7 @@
 <!--- your change affects other areas of the code, etc. -->
 
 ## Screenshots (if appropriate):
+
+## Changelog (neccesary)
+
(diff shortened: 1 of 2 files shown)


UndeadNappist:
After all roles are moved over and the transition period is smoothed out, all old roles are supposed to be deleted unless someone wishes to keep them for historical reasons.

The exceptions are Three Dog (it's already fallout), Host (I believe it plugs into other stuff so removing it would cause problems) and possibly admin (for the same reason).

I referenced this in the previous PR when warning John not to delete roles people were currently configured to because it would lead to headaches and problems.

improvedname:
Wrong PR nappist...

rhicora:
This seems like a good change, though some players may still want it in #dev-announcements. Thoughts on this @UndeadNappist ?

improvedname:
You can have both it doesn't hurt, it visualizes the PR in short, adds a update changelog in the game, and you can still to dev-announcements if you really want to.

Autismobirb:
Seems fine to me, Keeping #dev-announcements so head scribes can give there flare to it and annoy the ping whores.

improvedname:
https://cdn.discordapp.com/attachments/470355438572011530/560532469699575848/unknown.png
image

JohnOxford:
A valid plan would be to have development annoucements become the github bot's channel, as opposed to it being coding-public

then to merge this

GhostEcho:
Well part of dev announcements seems to be fun memery, though having a github bot channel would be nice, rather than it being in coding public

PR #724, opened 2019-03-27
Double the wait, double the fun!

GhostEcho:
Bumps respawn timer from 5 minutes to 10 minutes
Death now has double the consequences!

### DIFF
--- a/code/__DEFINES/mobs.dm
+++ b/code/__DEFINES/mobs.dm
@@ -257,4 +257,4 @@
 #define AI_VOX // Comment out if you don't want VOX to be enabled and have players download the voice sounds.
 
 //Respawn timer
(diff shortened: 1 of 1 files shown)


JohnOxford:
For future reference of this PR, a conversation was had with the following idea.

You can respawn as a member of the opposing faction [ Legion / NCR ] [ Vault / Brotherhood ] [ Den / Raider ] in five minutes, and no other faction, or you could wait fifteen minutes and spawn as a member of any other faction

This would however create a snowball but it's something i'll reference in the future, i completely agree with the bump in the time and OP screened many for their opinion and the consensus was 10 minutes.

Autismobirb:
This seems fine to me, bumping up the timer makes folk less likely to throw themselves into the frey with no fear of death because there just gonna respawn in the same amount of time it takes to make a cup o' tea.

rhicora:
This has been up for 24hrs and reviewed with +1 so I'll bring it in. I support it too.

PR #733, opened 2019-03-30
Changes to Stimpaks and Healing Powder

ChronicPwnage:
## Description

Buffs stimpacks, nerfs healing powder, differentiates the two to create distinct classes of healing compared to the current state of "stimpaks bad, healing powder good"

Before:
Stimpaks healed 40 burn/brute/oxy over 10 game ticks
Healing Powder healed 300 burn/brute over 100 game ticks

After:
Stimpaks initially heal 33.5 burn/brute/oxy (and 2.5 tox) over 5 game ticks, and reduce stuns during those 5 game ticks 
Stimpaks then heal about an additional 20 burn/brute/oxy over 28 game ticks (for a total of 53.5 brute/burn/oxy)
Healing Powder heals 186 burn/brute over about 62 game ticks (Heal at the same speed, around 2/3 total healing/duration so they don't last forever anymore)

Results:
Stimpaks heal very fast up front and then a little extra afterwards, and reduce stuns for a few seconds
Healing powder doesn't last a ridiculous amount of time anymore but is still good

## Motivation and Context

Healing powder is far, far better than stimpaks at the moment and is not very different. This change makes stimpaks (the harder to make healing source) more viable and differentiates the two in an interesting way.

## How Has This Been Tested?

Tested on a private server, everything worked fine.
## Changelog (neccesary)

🆑 ChronicPwnage
balance: Stimpaks now heal you a fair amount very quickly, then a little bit afterwards 
balance: Healing Powder total duration reduced by about 1/3
 /🆑

### DIFF
--- a/code/game/machinery/deployable.dm
+++ b/code/game/machinery/deployable.dm
@@ -52,9 +52,9 @@
 	else if(istype(mover, /obj/item/projectile))
 		if(!anchored)
 			return 1
-		//var/obj/item/projectile/proj = mover
-		//if(proj.firer && Adjacent(proj.firer))
-			//return 1
+		var/obj/item/projectile/proj = mover
+		if(proj.firer && Adjacent(proj.firer))
+			return 1
 		if(prob(proj_pass_rate))
 			return 1
 		/* /obj/item/restraints/legcuffs/bola/energy/throw_impact(atom/hit_atom)
--- a/code/modules/reagents/chemistry/reagents/medicine_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/medicine_reagents.dm
@@ -1304,7 +1304,7 @@
 	reagent_state = LIQUID
 	color = "#C8A5DC"
 	taste_description = "grossness"
-	metabolization_rate = REAGENTS_METABOLISM
+	metabolization_rate = 0.75 * REAGENTS_METABOLISM
 	overdose_threshold = 18
 
 /datum/reagent/medicine/stimpak_b/reaction_mob(mob/living/M, method=TOUCH, reac_volume, show_message = 1)
--- a/code/modules/reagents/chemistry/reagents/medicine_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/medicine_reagents.dm
@@ -1261,17 +1261,17 @@
 	..()
 	return TRUE
 
-/datum/reagent/medicine/stimpak
+/datum/reagent/medicine/stimpak_a
 	name = "Stimpak Fluid"
-	id = "stimpak"
-	description = "Rapidly heals damage when injected. Deals minor toxin damage if injested."
+	id = "stimpak_a"
+	description = "Rapidly heals damage when injected. Deals minor toxin damage if ingested."
 	reagent_state = LIQUID
 	color = "#C8A5DC"
 	taste_description = "grossness"
-	metabolization_rate = 3 * REAGENTS_METABOLISM
+	metabolization_rate = 5 * REAGENTS_METABOLISM
 	overdose_threshold = 20
 
-/datum/reagent/medicine/stimpak/reaction_mob(mob/living/M, method=TOUCH, reac_volume, show_message = 1)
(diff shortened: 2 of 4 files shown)


improvedname:
Bad solution, keep it in orginal form but with stimpak, metabolism lowered from 3 to 0.75

Light-Weave:
@improvedname IIRC that'll mean stimpacks still heal the same amount, but slower. If you're increasing the duration, increase the heal as well, since the issue here is that healing powder has a massive heal for a long time.

improvedname:
You don't need to increase the heal, or create more chems to do so, the only difference between healing powder should vs stimpack is like applt time vs stimpaks, stimpaks are instant. just decrease the stimpaks metabolism, which creates the true weakness of the healing. it had about 3-4 ticks because of it including like 10 units so barely making up for a 40 damage heal. so just decrease the orginal metabolism.

ChronicPwnage:
We discussed this over Discord but to summarize my points, I think healing powder and stimpaks should be differentiated in an interesting way with stimpaks being a different upgrade instead of just a straightforward upgrade. I think healing powder lasts far too long right now. I understand your reasoning for your proposed changes in #734 but I don't agree with the direction they take the game personally, I think this way is more interesting. I would love to hear more input from others about how they believe stimpaks/powder should function.

code/modules/reagents/reagent_containers/hypospray.dm
```diff
@@ -179,10 +179,10 @@
 	name = "stimpak"
 	desc = "A handheld delivery system for medicine, used to rapidly heal physical damage to the body."
 	icon_state = "stimpakpen"
-	volume = 10
-	amount_per_transfer_from_this = 10
-	list_reagents = list("stimpak" = 10)
-	
+	volume = 20
+	amount_per_transfer_from_this = 20
+	list_reagents = list("stimpak_a" = 10, "stimpak_b" = 10)
```
UndeadNappist: BUT WHY.

code/modules/reagents/chemistry/recipes/medicine.dm
```diff
@@ -256,6 +256,6 @@
 /datum/chemical_reaction/stimpak
 	name = "Stimpak Fluid"
 	id = "stimpak"
-	results = list("stimpak" = 2)
+	results = list("stimpak_b" = 2)
```
UndeadNappist: but why

code/modules/reagents/chemistry/reagents/medicine_reagents.dm
```diff
@@ -1261,34 +1261,70 @@
 	..()
 	return TRUE
 
-/datum/reagent/medicine/stimpak
+/datum/reagent/medicine/stimpak_a
```
UndeadNappist: Why are there two types of stimpak fluid now.

code/game/machinery/deployable.dm
```diff
@@ -52,9 +52,9 @@
...
 			return 1
-		//var/obj/item/projectile/proj = mover
-		//if(proj.firer && Adjacent(proj.firer))
-			//return 1
+		var/obj/item/projectile/proj = mover
+		if(proj.firer && Adjacent(proj.firer))
+			return 1
 		if(prob(proj_pass_rate))
 			return 1
 		/* /obj/item/restraints/legcuffs/bola/energy/throw_impact(atom/hit_atom)
```
UndeadNappist: What does this have to do with anything?

UndeadNappist:
This is not the appropriate way to make this object. If you want stimpak injectors blended by default for didfferent healing rates, use stimpak fluid + tricordizaine or some already pre-existing chemical.

This adds unnecessary code bloat, is very spaghetti, and quite strange - and would likely lead to a vast number of problems down the line.

Similarly, from a design perspective, I outright stated I wanted a **clear difference** between chemicals to make each one **distinct** and their drawbacks and advantages giving a wide variety.

Inappropriate from a code perspective, and inappropriate from a design perspective.

PR #737, opened 2019-03-31
c45 Sanitization

GhostEcho:
Comments out the c45 projectile strings in nukie smg code to prevent issues that we had before


UndeadNappist:
Won't this runtime when it tries to call on a nonexistent projectile for guns?

GhostEcho:
gonna close and re-open a PR since this branch disappeared in my github

PR #745, opened 2019-04-01
Increase of 50% to respawn time

improvedname:
## Changelog (neccesary)
:cl: Improvedname
tweak: Changes respawn timer from 10 to 15
:cl:

### DIFF
--- a/code/__DEFINES/mobs.dm
+++ b/code/__DEFINES/mobs.dm
@@ -257,4 +257,4 @@
 #define AI_VOX // Comment out if you don't want VOX to be enabled and have players download the voice sounds.
 
 //Respawn timer
-#define RESPAWN_TIMER 6000 // in deciseconds, the time to wait before allowing respawn. 10 minutes = 10 * 60 (seconds in minute) * 10 (seconds to deciseconds)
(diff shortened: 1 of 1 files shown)


JohnOxford:
I strongly disagree on moving it past ten minutes, and here's why.

The reason that we originally bumped it up from five minutes to ten minutes was to solve the problem regarding people not valuing their life, abusing respawns, and generally facilitating a deathmatch-esqe server. 

The argument that was had when this occurred was that no one wanted to wait a really long time and that they would just leave before they took the time to wait and rejoin the game. This is heavily tied into maintaining a stable population. The game is intrinsically less fun to play with significantly less people on the map.

So in order to compromise, we agreed to find a balance between waiting long enough to let a player cool down from any conflicts that they were in, to discourage abuse of respawn, and to make players value their life, but not to bore them to the point where they have enough time to go find something else to do, by closing out of the server, or going and joining another server.

My firm opinion is that we either leave it at ten or disable respawn completely and fall deeper into the rabbit hole that is HRP F13.

UndeadNappist:
Updated information: it's currently at ten but the server's respawn timer is at five. The new number has not been deployed.

PR #759, opened 2019-04-03
Who the hell thought this was a good idea

GhostEcho:
Alright, there needs to be a discussion on how much shit people throw into one PR

### DIFF
--- a/code/modules/crafting/recipes.dm
+++ b/code/modules/crafting/recipes.dm
@@ -297,7 +297,7 @@
 	category = CAT_WEAPONRY
 	subcategory = CAT_AMMO
 
-/datum/crafting_recipe/c4570
+/*/datum/crafting_recipe/c4570
 	name = ".45-70 speed loader (NCR)"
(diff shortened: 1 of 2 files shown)


Light-Weave:
1) commenting the speedloader out is gonna break shit because the speedloaders actually existed already
and 2) the sequoia does terrible damage anyway because -20 AP, just increase the cost or something- right now it's worth 1.5x a .44 speedloader but we could make it 2x (plus the 2x multiplier from the ammobench being inefficient) = 8-10 sheets 

having uncraftable ammo is dumb when the ammo sucks

JTGSZ:
Hello, for some future advice try using ctrl +f.
Untitled

Light-Weave:
update: we came to a SUPER COOL conclusion in the discord (craftable .45-70 lowpower) so 
take this PR down and I'll fix the recipe in the upcoming 4570LP PR

GhostEcho:
Indeed @Light-Weave talk with Ren about it though and work with him for numbers

PR #772, opened 2019-04-05
Bolas

Cw3040:
:cl:
fix:legion get normal bolas instead of reinforced meme ones
/ :cl:

### DIFF
--- a/code/game/objects/items/handcuffs.dm
+++ b/code/game/objects/items/handcuffs.dm
@@ -320,7 +320,7 @@
 	name = "bola"
 	desc = "A restraining device designed to be thrown at the target. Upon connecting with said target, it will wrap around their legs, making it difficult for them to move quickly."
 	icon_state = "bola"
-	breakouttime = 35//easy to apply, easy to break out of
+	breakouttime = 60//easy to apply, easy to break out of
 	gender = NEUTER
 	var/knockdown = 0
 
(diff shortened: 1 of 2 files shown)


improvedname:
Increase the resist time of thr normql ones

Cw3040:
Done :^)

Light-Weave:
Did you change the ones in the armory too?

Cw3040:
None start in the armory.

code/game/objects/items/handcuffs.dm
```diff
@@ -320,7 +320,7 @@
 	name = "bola"
 	desc = "A restraining device designed to be thrown at the target. Upon connecting with said target, it will wrap around their legs, making it difficult for them to move quickly."
 	icon_state = "bola"
-	breakouttime = 35//easy to apply, easy to break out of
+	breakouttime = 60//easy to apply, easy to break out of
```
UndeadNappist: Doing this will make them have the same net effect, except now raiders have access to it since they can just make their own. Slowdowns are dangerous in combat.

Light-Weave: They should be- but knockdowns are far, far worse. Not being able to instantly remove it means that any conflict with a bola will likely become a melee fight.

GhostEcho: Changed to 4.5 seconds, feels good to me but we'll see in practice

Issue #738, opened 2019-03-31
Balance Issue with Miser Gutsy


Jerkmilder:
I'm seeing a bit of a balance problem, how are we an RP server if we have NPC's like Mr. Gutsy. As soon as they walk into  #town they kill anyone traveling alone, quite literally instantly, Can't run away due to them running faster than you if they hit you once with their gun, and they're a tiny bit tanky. NPC's like these either should not exist or should rather be nerfed down a tad.
I can't see how this can be anything similar to an RP server with escalation if an instant kill NPC is even allowed anymore. "Oh dying's apart of the game, you can die just instantly and remake!" Yes, I know, but it shouldn't be like that for NPC's, they shouldn't be instantly killing you because someone wished to move them out into the public so they can kill 5-6 people.

Potential Fixes- 
Make it easier to defib people, by extending the defib timer just a tad more, maybe a couple more minutes, because fighting these things alone are quite literally, stupid as hell. (Unintentionally fighting them alone that is.)

Just remove him from the dungeon that he spawns in, Apparently, it's NORMAL for him to just kill 5-6 people around due to people kiting him into town, replace it with other creatures.

Or give him pathing, where he doesn't really leave the area he's supposed too.

JTGSZ:
Hello, the wasteland mobs want you to die. So mister gutsy turning people into corpses is a good thing.

JTGSZ:
Also im going to just close this

Light-Weave:
Gutsy has a 3-round burst, with each round doing 35 damage (or realistically ~20 damage because it has AP of -18). It also has 200 health and 72 damage melee attacks. If a Gutsy lands even one burst on someone, that someone is dead. Not so with the protectrons, which do a maximum of 60 damage per burst (20x3) and do only 25 melee damage.

TL;DR: Nah, Gutsy actually kind of is OP. Drop the 3-round burst to 1 and it'd still work, because its AI forces it to close to melee range for the devastating 72-damage flamer- the burst gun is a supplement.

PR #785, opened 2019-04-06
Plasma

Cw3040:
:cl:
edit: /obj/item/projectile/plasma/pistol improved, damage increased by 15 and AP value removed
edit: /mob/living/simple_animal/hostile/handy/gutsy uses /obj/item/projectile/plasma/scatter now, a 25 damage AP 0 projectile
/:cl:


Light-Weave:
This also means that Gutsy is less effective against unarmored targets and more effective against armored ones (i.e. more consistent), thereby fixing #738 - it can still fuck you up (and is a threat even against armored enemies), but it can't one-shot crit an unarmored wastelander now.

PR #786, opened 2019-04-07
Fixes ticklag/FPS

JohnOxford:
This was tested, a visual comparison was made between 0 (syncs with tick rate) 40, 60, and 120 FPS. I posted this in the coding channel in the discord. 

What i imagine happened was setting the FPS to zero in the config file doesn't sync it to the tick rate, as opposed to setting it to zero in game preferences.


Light-Weave:
ok but did you test it this time

PR #794, opened 2019-04-08
Update vote.dm

ChronicPwnage:
## Description

When admins aren't on, it changes the one minute rebooting world after a successful vote to a 20 minute shuttle call
## Motivation and Context

A lot of people vote restart thinking they'll have 20 minutes left in the round after a successful vote, and people are disappointed when it turns out "moderators" don't count as admins and the world reboots after 1 minute. This fixes that.
## How Has This Been Tested?

Local machine, worked fine.
## Changelog (neccesary)
:cl:
tweak: A successful restart vote now calls the shuttle instead of rebooting the world after 1 minute
/:cl:

### DIFF
--- a/code/controllers/subsystem/vote.dm
+++ b/code/controllers/subsystem/vote.dm
@@ -131,7 +131,8 @@ SUBSYSTEM_DEF(vote)
 				active_admins = 1
 				break
 		if(!active_admins)
-			SSticker.Reboot("Restart vote successful.", "restart vote")
+			SSshuttle.emergency.request()
+			//SSticker.Reboot("Restart vote successful.", "restart vote")
 		else
 			to_chat(world, "<span style='boldannounce'>Notice:Restart vote will not restart the server automatically because there are active admins on.</span>")
 			message_admins("A restart vote has passed, but there are active admins on with +ban, so it has been canceled. If you wish, you may restart the server.")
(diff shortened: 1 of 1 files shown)


improvedname:
The last time i did this, players voted restart again, it passed and the shuttle timer got reset so add a check for that.

ChronicPwnage:
> 
> 
> The last time i did this, players voted restart again, it passed and the shuttle timer got reset so add a check for that.

Nice catch, I've changed it to disable restart voting upon a successful restart vote.

GhostEcho:
So this seems to have broken restart votes, as most people cant call one, including coder ranked members, only an admin can

code/controllers/subsystem/vote.dm
```diff
@@ -280,6 +281,7 @@ SUBSYSTEM_DEF(vote)
 			if(CONFIG_GET(flag/allow_vote_restart) || usr.client.holder)
 				if(min_restart_time < world.time)
 					initiate_vote("restart",usr.key)
+					CONFIG_SET(flag/allow_vote_restart, FALSE)
```
GhostEcho: I think its this line

Cw3040: It's a feature.

ChronicPwnage: It looks like configs are saved between rounds, and I didn't catch it because on a private server you automatically have full privileges. Forcing voting to be enabled on server restart should fix this, and have the added benefit of making it less likely that an admin will log off leaving votes disabled by accident.

PR #796, opened 2019-04-08
Minor Bugfixes

Light-Weave:
## Description
If you spawn with prescription glasses and your job already has glasses equipped, those glasses are deleted, but never officially removed- so if it's something like a medHUD or NVGs, you still get the benefits of both. Now you start with prescription glasses in your hands if your eyes are already full.

The job select screen and manifest screen has been rearranged slightly to accommodate for the new roles.

## Motivation and Context
Bugfixes.

## How Has This Been Tested?
Yup.

## Changelog (neccesary)
:cl:
fix: Fixes job select screen and manifest screen.
fix: Fixes minor prescription glasses exploit.
/:cl:

### DIFF
--- a/code/modules/jobs/jobs.dm
+++ b/code/modules/jobs/jobs.dm
@@ -1,8 +1,12 @@
 GLOBAL_LIST_INIT(command_positions, list(
 	"Elder",
+	"Head Scribe",
+	"Paladin",
 	"Centurion",
 	"NCR Captain",
+	"NCR Veteran Ranger",
 	"Overseer",
+	"Chief of Security",
 	"Sheriff"
 ))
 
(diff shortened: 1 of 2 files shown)


code/modules/jobs/jobs.dm
```diff
@@ -18,25 +22,31 @@ GLOBAL_LIST_INIT(brotherhood_positions, list(
...
+	"Legion Veteran Decanus",
 	"Veteran Legionnaire",
-	"Legionary",
+	"Legion Prime Decanus",
+	"Prime Legionnaire",
+	"Legion Recruit Decanus",
+	"Recruit Legionnaire",
+	"Legion Vexillarius",
+	"Legion Explorer",
+	"Legion Scout",
```
UndeadNappist: These are a bunch of changes without changing the actual job titles, and will therefore probably break things. Last time we did something like this people ended up spawning in the arrivals shuttle and other wacky things.

UndeadNappist:
Generally I disapproved of adding medical doctors, but on the flip side, the NCR are super bland and boring.

code/modules/jobs/jobs.dm
Light-Weave: I *actually tested it* and it works fine tbh, these are literally just for the manifest/job selection screen

PR #798, opened 2019-04-08
Simplemob Armor Framework

Cw3040:
Name says it all, adds armor values to simplemobs that will handle
incoming damage in addition to the existing damage_coefficient

:cl:
add: Adds armor values for simplemobs, default is 0 to not make existing mobs OP
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/simple_animal.dm
+++ b/code/modules/mob/living/simple_animal/simple_animal.dm
@@ -48,6 +48,7 @@
 	var/armour_penetration = 0 //How much armour they ignore, as a flat reduction from the targets armour value
 	var/melee_damage_type = BRUTE //Damage type of a simple mob's melee attack, should it do damage.
 	var/list/damage_coeff = list(BRUTE = 1, BURN = 1, TOX = 1, CLONE = 1, STAMINA = 0, OXY = 1) // 1 for full damage , 0 for none , -1 for 1:1 heal from that source
+	var/list/armor = list(BRUTE = 0, BURN = 0, TOX = 0, CLONE = 0, STAMINA = 0, OXY = 0)
 	var/attacktext = "attacks"
 	var/attack_sound = null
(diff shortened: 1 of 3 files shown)


code/modules/mob/living/simple_animal/simple_animal.dm
```diff
@@ -48,6 +48,7 @@
 	var/armour_penetration = 0 //How much armour they ignore, as a flat reduction from the targets armour value
 	var/melee_damage_type = BRUTE //Damage type of a simple mob's melee attack, should it do damage.
 	var/list/damage_coeff = list(BRUTE = 1, BURN = 1, TOX = 1, CLONE = 1, STAMINA = 0, OXY = 1) // 1 for full damage , 0 for none , -1 for 1:1 heal from that source
+	var/list/armor = list(BRUTE = 0, BURN = 0, TOX = 0, CLONE = 0, STAMINA = 0, OXY = 0)
```
UndeadNappist: Doesn't this make them invulnerable?

Cw3040: No, those are percentile values. You're thinking damage_coeff which does in fact make them invulnerable when set to 0

Cw3040:
If this is actually merged, it will be as a harbinger of a full wasteland mob rework in the same vein as the most recent armor rework. It'll make armor-piercing weapons more effective against simplemobs and non-armor-piercing weapons slightly less so.

LetterN:
because we dont want to fucking waste important lists shits on simplemobs

UndeadNappist:
> because we dont want to fucking waste important lists shits on simplemobs

LAZY* handles that and optimally it wouldn't change anything except organize stuff.

RenGusta:
giving this a hard close for now, until it's brought up again and spoken about further. Interesting concept, little out of date now.

PR #817, opened 2019-04-13
Adds cover to dense objects

Light-Weave:
## Description
Dense objects now no longer act as guaranteed portable cover- instead, they only block 35% of projectiles. If you're adjacent to them, they don't block any, like sandbags, so you can use them as really, really ghetto barricades if you want. (Don't expect them to save you, though.)

Similarly, fences now block 40% of projectiles, unless adjacent. This lets you use the Legion base entrance as a giant deathtrap, if you're so inclined- you can stand behind the fence and gun people down while taking limited fire in return.

This does **not** apply to people or turfs, so you can't shoot through walls... unless those walls are girders, in which case you can, because girders are actually fairly good cover- doubly so for reinforced girders.

Pass rate = how many projectiles ignore the cover. Lower is stronger cover.
**Machinery, computers, frames, structures, closets, crates and displaced girders**: 65% pass rate
^this covers pretty much everything you can build and move around.
**Fence barricades**: 50% pass rate
^only 50% because fences themselves have a 40%, and it checks for both (bringing it to 20% combined).
**Fences**: 40% pass rate
**Barricades**: 35% pass rate
**Sandbags**: 20% pass rate
**Reinforced girders**: 20% pass rate

## Motivation and Context
You now can't cheese the enclave bunker with 50 sheets of metal and guaranteed portable cover. Buffing fences is also a defensive boost anywhere fences can be found- but if an attacker can get up close to the fence, the tables can turn.

## How Has This Been Tested?
Yup.

## Changelog (neccesary)
:cl:
balance: Dense objects (e.g. fences, computers, machine frames, closets) can now be shot through while adjacent.
balance: While not adjacent, dense objects provide varying amounts of cover depending on the object in question. Moveable objects such as closets provide 35% cover, while static ones such as reinforced girders provide more.
/:cl:

### DIFF
--- a/code/game/machinery/_machinery.dm
+++ b/code/game/machinery/_machinery.dm
@@ -94,6 +94,8 @@ Class Procs:
 	var/obj/item/circuitboard/circuit // Circuit to be created and inserted when the machinery is created
 	var/interaction_flags_machine = INTERACT_MACHINE_WIRES_IF_OPEN | INTERACT_MACHINE_ALLOW_SILICON | INTERACT_MACHINE_OPEN_SILICON | INTERACT_MACHINE_SET_MACHINE
 	var/machine_tool_behaviour = NONE //can it be used as a tool in crafting?
+	var/barricade = 1 //if 1, acts as barricade
+	var/proj_pass_rate = 80 //percentage change for bullets to fly over, if barricade=1
 
 /obj/machinery/Initialize()
 	if(!armor)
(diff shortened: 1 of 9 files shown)


Light-Weave:
travis is still broken, it's fine

improvedname:
So tell me why i shouldnt just use girders/computers/machinery

Light-Weave:
because girders already let bullets through and if you use computers/machinery we'll fuck with those too

improvedname:
do it in one go. because that shall happen

Light-Weave:
COVER OVERHAUL GO GO GO

Light-Weave:
wall-mounted shit is fine because it checks for density anyway and wall-mounted shit isn't dense, and nondense stuff Cannot act as a barricade (if you can move through it, so can bullets!)

girders have their own shit about bullets still but this just means you can use them to hide behind

and it exists because people are cheesing dungeons hardcore with guaranteed cover, and "you can shoot it open if you magdump hard enough" doesn't really matter when you can build like 25 lockers with one stack of metal or just scavenge a bunch of crates to bring with you- taking cover and being able to shoot over it (and be shot in return) is way more immersive and awesome

monkeys-on-chairs is an exploit but it's out of scope for this PR given that it'd interfere with shooting actual people on chairs too

ChronicPwnage:
I agree that most shit shouldn't block 100% of shots until it's destroyed.
This is a good PR, good job!

PR #827, opened 2019-04-15
Modifies shooting delay (no more paralysis) and adds a throwing delay

ChronicPwnage:
## Description

This PR adds two new status effects (like stunned, weakened, paralyzed etc) that function to add a delay when you're currently drawing a firearm or throwing a throwing weapon. If you're affected by either status effect, you can still walk around, equip stuff, drop stuff, etc. 
his means no more total body paralysis when drawing a firearm, and it makes these things easy to modify with stuff like holsters, drugs, etc.
## Motivation and Context

Weapon draw delay: No one likes the fact that you're totally paralyzed when you draw a gun, it feels unnatural not to be able to move while doing it. This change makes it possible to move and draw at the same time, and makes it easy to add things like quick-draw holsters and the like in the future.

Throwing delay: In its current state, if you're quick with hotkeys, you can machinegun out a quiver of throwing spears faster than a fucking L6 saw and break down a door/window/person just as fast. It's hilarious to watch but ultimately super broken and needs a fix. This makes it so you can't do that. Currently it's only applied to throwing weapons, since other things don't usually do that much damage, but could be applied to other things as well if we see abuse of mechanics in that way.
## How Has This Been Tested?

I've tested the changes on my personal machine and fixed the bugs that popped up. I did notice another issue, but after testing it on the live server I realized that it's not my code doing it and will look into fixing it as well (You can't fire a gun if you click right next to you, you have to fire beyond. Something to do with point blank code I presume)
Screenshot
## Changelog (neccesary)
:cl: ChronicPwnage:
add: Adds a weapon draw delay, preventing you from firing a weapon right after you draw it
add: Adds a throwing weapon delay, preventing machine gun throwing spears from YATATATAT-ing doors (and people) in moments
del: Removes the existing weapon draw paralysis, which paralyzed you entirely when you drew a weapon
/:cl:

### DIFF
--- a/code/__DEFINES/status_effects.dm
+++ b/code/__DEFINES/status_effects.dm
@@ -47,6 +47,10 @@
 
 #define STATUS_EFFECT_SLEEPING /datum/status_effect/incapacitating/sleeping //the affected is asleep
 
+#define STATUS_EFFECT_WEAPON_DRAW_DELAYED /datum/status_effect/incapacitating/weapon_draw_delayed //The affected can't fire a weapon
+
+#define STATUS_EFFECT_THROW_DELAYED /datum/status_effect/incapacitating/throw_delayed //The affected can't throw something
+
 #define STATUS_EFFECT_BELLIGERENT /datum/status_effect/belligerent //forces the affected to walk, doing damage if they try to run
 
 #define STATUS_EFFECT_GEISTRACKER /datum/status_effect/geis_tracker //if you're using geis, this tracks that and keeps you from using scripture
--- a/code/datums/status_effects/debuffs.dm
+++ b/code/datums/status_effects/debuffs.dm
@@ -37,6 +37,14 @@
 	if(owner.getStaminaLoss())
 		owner.adjustStaminaLoss(-0.3) //reduce stamina loss by 0.3 per tick, 6 per 2 seconds
 
+//WEAPON DRAW DELAYED
+/datum/status_effect/incapacitating/weapon_draw_delayed
+	id = "weapon_draw_delayed"
+
+//THROW DELAYED
+/datum/status_effect/incapacitating/throw_delayed
+	id = "throw_delayed"
+
 
 //UNCONSCIOUS
 /datum/status_effect/incapacitating/unconscious
--- a/code/game/objects/items.dm
+++ b/code/game/objects/items.dm
@@ -791,7 +791,6 @@ GLOBAL_VAR_INIT(rpg_loot_items, FALSE)
 /obj/item/gun/proc/weapondraw(obj/item/gun/G, mob/living/user) // Eventually, this will be /obj/item/weapon and guns will be /obj/item/weapon/gun/etc. SOON.tm
 	user.visible_message("<span class='danger'>[user] grabs \a [G]!</span>") // probably could code in differences as to where you're picking it up from and so forth. later.
 	user.SetWeaponDrawDelay(max((4 * G.weapon_weight + 1),(user.AmountWeaponDrawDelay())))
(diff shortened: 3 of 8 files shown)


SpaceVampire:
Please never add quick-draw holsters, the current stun is great for dealing with escalation issues, as it gives people a moment to react.

ChronicPwnage:
> 
> 
> Please never add quick-draw holsters, the current stun is great for dealing with escalation issues, as it gives people a moment to react.

I think it's definitely something that requires a lot of discussion if it was ever added, but I was just using it as an example of ways that the draw speed could be affected by having it coded this way (just like stuns and knockouts are affected by stuff like drugs) - we could also do the opposite for example a backpack that can hold full-size guns (for people who want to sell guns) but there's a large drawspeed delay.

ChronicPwnage:
> 
> 
> Modify it to be dynamic instead of flat for the weapon draw delay and derived slowdown var just in case other things somehow get applied to slowdown, etc.
> 
> Pretty simple to do for slowdown, it's just
> (src.slowdown + etc)

I'm not entirely sure what you mean. Do you mean change:
`user.SetWeaponDrawDelay(4 * G.weapon_weight + 1)` ?

If I understand you correctly (I am not sure that I do, pretty tired right now) the issue with that would be that it would keep adding a delay that added up if someone keeps drawing/holstering a weapon. I guess a solution would be to do `user.SetWeaponDrawDelay(max(AmountWeaponDrawDelay(), 4 * G.weapon_weight + 1))` ?

What do you think?

UndeadNappist:
> I'm not entirely sure what you mean. Do you mean change:
> `user.SetWeaponDrawDelay(4 * G.weapon_weight + 1)` ?
> 

SetWeaponDrawDelay(((4 * G.Weapon_weight + 1) + CurrentWeaponDrawDelay))

You can name them whatever you want, but it's basically taking whatever your current weapon draw delay is and then modifying that, instead of flat-setting the weapon draw delay.

As you can see in my derived slowdown proc, it's wrong there as well ; it sets the slowdown without respect to other factors that might affect the slowdown (including the qualities of the weapon itself).

The reason to do this is if we want to go further into it and tinker with other slowdowns or interesting interactive effects later. Either way, it's more dynamic.

ChronicPwnage:
Okay well I'm still not really getting what you mean.

Right now you can affect the weapon draw delay directly on the mob that draws the weapon, just like you affect stuns. For example, adding a chemical "speedydrawnium" that moves your weapon draw delay towards 0 every tick in your bloodstream. Or adding a gun that reduces your weapon draw delay by 2 when you grab it (so it'd calculate the `4 * G.weapon_weight + 1` and then reduce that by 2)

I mean I could make it easier by adding a variable to weapons like draw_time and incorporating that to the formula, so you can change it/affect the weapon more directly. But like you said, your example of `SetWeaponDrawDelay(((4 * G.Weapon_weight + 1) + CurrentWeaponDrawDelay))` would just keep increasing the draw delay.

UndeadNappist:
And what happens when they've drawn a long gun, have a certain weapon delay, and then immediately draw a smaller weight pistol?

Does their draw delay go __down?__

ChronicPwnage:
> 
> 
> And what happens when they've drawn a long gun, have a certain weapon delay, and then immediately draw a smaller weight pistol?
> 
> Does their draw delay go **down?**

So you're worried about someone drawing a long gun, switching hands, drawing a pistol, dropping the pistol, and switching back, having only the pistol draw time? Honestly the amount of time it'd take to do that considering lag even with a macro wouldn't be enough to affect gameplay without having some gun with an absolutely ridiculous draw time. But I get your point, and can see a simple solution.

PR #832, opened 2019-04-16
unfucks travis: webedit edition

LetterN:
yes


AffectedArc07:
I swear if you make one more web edit you get the basement for a month straight

PR #839, opened 2019-04-17
fixes rnd console runtime

LetterN:


LetterN:
we wont be needing this

PR #842, opened 2019-04-17
moves all technopreak shit to a trait

### LINKED ISSUE #838: Runtime in null.mind
[Priority]: # Low

[PR]: # Mine, having do with moving technophreak away from a quality and into a mind datum.

Basically what's happening is the paperdoll preview and mid-transit results in it calling null because the mob has no mind to attach technophreak to, and similarly, calls null when force-equipping because job code is a pile of junk and so are the equip codes.

Oddly, when I fixed the null call it still worked for the paperdoll preview but __not__ for the transit ; character job would spawn in with no power armor equipped.

Tried various workarounds including for(!mind) and other stuff, and after six hours basically gave up. Good luck out there.

LetterN:
might fix #838


GhostEcho:
If it proves to be an issue we'll change it later

PR #863, opened 2019-04-22
Slightly Less Bad Deathclaw

Light-Weave:
## Description
Deathclaws now no longer have resistance to small-arms fire, because y'all are unrobust.

## Motivation and Context
Per request.

## How Has This Been Tested?
Ayup.

## Changelog (neccesary)
:cl:
balance: Turns out deathclaws are actually vulnerable to pistols after all.
/:cl:


GhostEcho:
Im actually going to bump this to 15% for now

PR #875, opened 2019-04-27
Removes curses from masks

Light-Weave:
## Description
Cursed masks are dumb. They're no longer cursed to be nodrop.

## Motivation and Context
Cause they're dumb and cause problems.

## How Has This Been Tested?
Ayup.

## Changelog (neccesary)
:cl:
tweak: Cursed masks are no longer nodrop.
/:cl:

### DIFF
--- a/code/modules/clothing/masks/miscellaneous.dm
+++ b/code/modules/clothing/masks/miscellaneous.dm
@@ -68,7 +68,7 @@
 	icon_state = "joy"
 
 /obj/item/clothing/mask/joy/joyful
-	item_flags = NODROP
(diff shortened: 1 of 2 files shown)


Light-Weave:
Punished Raider doesn't have default job slots, so I'm not too fussed about touching it (especially given that historically touching job datums does not end well) since they only spawn with safe, non-cursed masks now. None of that nodrop business.

rhicora:
Makes sense, I'll merge and add you as Scribe.

pireamaineach:
Forgot the horsemask.

pireamaineach:
We had a fiasco with the horse mask today. Had to cut off my head then re-attach it, two arms aswell before it was finally burned.

PR #878, opened 2019-04-30
Powerfist

GhostEcho:
Okay, the one hit memes were fun, but this is a bit much now

Powerfists ignore armor, so this will still be a powerful tool, but not the one tapper it has been in the past


Light-Weave:
travis shush

PR #885, opened 2019-05-11
Replacing default SS13 Cyborgs with Mr. Handy models

SaintWaIker:
## Description
Replaces the default models of all cyborgs, and their lit/unlit eyes, save for Maid, Butler, and Brobot sub-selections in the Service category. SS13 models are fine and all, but the immersion of having various Mr. Handy cyborgs is worth the change. Removed most transformation animations by replacing the file with a still image of a standard Mr. Handy, to save on filesize and immersion breaking. 

## Motivation and Context
Autism and IMMERSION

## How Has This Been Tested?
Compiled and hosted on my machine. Found misalignment with the cyborg eyes when unlit and lit due to model variations, but that has been corrected.

## Screenshots:
https://gyazo.com/e1fcf9c8df0a55e61abf0634f0e0006d
## Changelog
imageadd: added some icons and images
imagedel: deleted some icons and images
code: changed some code


SaintWaIker:
Making the module name Tophat (Handy) caused coding issues, should pass now that I removed the parentheses.

PR #886, opened 2019-05-11
Stun balance

Rohesie:
## Changelog (neccesary)
:cl:
balance: Soap stun time reduced from 8 to 3 seconds.
balance: Regular bolas take 6 instead of 4.5 seconds to be removed. No stun, hard to remove.
balance: Reinforced bolas stun for 1 second instead of 2, and take 3 seconds to be removed instead of 7. Short stun, easy to remove, hopefully viable now.
balance: All slippery plants, like bananas, have their slippery gene power halved, so will also suffer shorter stuns. A banana can reach up to 7 seconds of stun at maximum potency.
/:cl:

### DIFF
--- a/code/game/objects/items/clown_items.dm
+++ b/code/game/objects/items/clown_items.dm
@@ -29,7 +29,7 @@
 
 /obj/item/soap/Initialize()
 	. = ..()
-	AddComponent(/datum/component/slippery, 80)
+	AddComponent(/datum/component/slippery, 3 SECONDS)
 
 /obj/item/soap/nanotrasen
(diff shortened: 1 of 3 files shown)


Light-Weave:
Not huge on the r-bola stun, even if it only lasts a second, because even when you get up you still can't fire until the fire-delay wears off. Maybe just have it increase the slowdown even more (so with a r-bola, you basically can't move at all)?

RenGusta:
Discussed with me at length at its inception, and having a short stun and quickly removable bola was deemed balanced, alongside its scarcity.

PR #894, opened 2019-05-15
Makes stim pills realistic

GhostEcho:
Stimulant pills are now more realistic, similar to the ones you find in stores


improvedname:
its called how do you make stim packs useless

PR #914, opened 2019-05-26
Restricts interactions to humans like before (funclaws still work)

Digdugxx:
The funclaw PR unintentionally enabled erp with all mobs. This restricts that again. Note funclaws still work

### DIFF
--- a/interactions/interaction_interface.dm
+++ b/interactions/interaction_interface.dm
@@ -14,6 +14,8 @@
 		return
 	if(usr.restrained())
 		return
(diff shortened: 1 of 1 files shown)


SpaceVampire:
That sounds like a feature. Erika Wagner is gone now, whats the worst that could happen?

pireamaineach:
I mean if we have a deathclaw that fucks you why not let people get murdered for fucking brahmin?

pireamaineach:
> sure

Why have you done this to us.

PR #950, opened 2019-06-03
drops powerfist force to 45

AdamElTablawy:
OOC: Adam El-Tablawy: I can open a PR right now to drop the fucking powerfist's force to 30 if I felt like it

OOC: Snooper44: Okay retard do it.
## Description

powerfist strong, apparently, and people are mad that at max setting it literally instant kills given that it does a staggering 216 damage

while they're incapable of using basic reasoning to figure out that a shotgun dart is just as powerful i hate big number so now it does 90 damage on force setting 2 and 135 on force setting 3, enough to solidly drop someone into crit (not accounting for armor) but not enough for an instant kill

### DIFF
--- a/code/game/objects/items/powerfist.dm
+++ b/code/game/objects/items/powerfist.dm
@@ -7,7 +7,7 @@
 	righthand_file = 'icons/mob/inhands/weapons/melee_righthand.dmi'
 	flags_1 = CONDUCT_1
 	attack_verb = list("whacked", "fisted", "power-punched")
-	force = 72
+	force = 45
 	throwforce = 10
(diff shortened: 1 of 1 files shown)


AdamElTablawy:
oh my god the fucking thing DIRECTLY APPLIES DAMAGE AND BYPASSES ALL ARMOR

target.apply_damage(force * fisto_setting, BRUTE)

LOOK AT THIS

WHAT THE FUCK WHO CODED THIS

Rohesie:
Might be more interesting to fix that bug before tweaking the values.

PR #961, opened 2019-06-06
nerfs powerfist into the ground

AdamElTablawy:
discussed with ren

30 damage

60 AP

go fuck yourselves

penetrates armor, 4 hit crit in basically all situations, good against heavily armored, bad against low armor by virtue of being overkill and unfavorable compared to chainsaw in those

requires no gas, does not throw, no oxygen tanks, nada

fits in backpacks so it's not just dead weight. functionally this is an armor penetrating machete

## Changelog (neccesary)
:cl:
balance: Powerfist now the same damage as a machete and penetrates all but the heaviest melee armor. Does not throw hit people, does not stun by knocking them into walls, requires no oxygen tanks to use, and fits in backpacks.
/:cl:

### DIFF
--- a/code/game/objects/items/powerfist.dm
+++ b/code/game/objects/items/powerfist.dm
@@ -11,6 +11,6 @@
 	armour_penetration = 60
 	throwforce = 10
 	throw_range = 7
-	w_class = WEIGHT_CLASS_NORMAL
+	w_class = WEIGHT_CLASS_BULKY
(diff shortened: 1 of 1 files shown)


AdamElTablawy:
nevermind a short OOC conversation has convinced me that the powerfist should not be able to fit in backpacks

everyone thank nolafregit

AdamElTablawy:
merge pls

PR #967, opened 2019-06-07
Replaces default metal wall sprite with a new corrugated metal wall sprite.

GoodG1448:
I personally dislike the look of the default metal wall in this setting, I decided to make it look like corrugated metal since it fits with the post-nuclear aesthetic.

The walls look like this in game:
reference final


GhostEcho:
Are we Eris now? Think they could use a little more work, seems a little off

GoodG1448:
Blame it on how the wall sprites link with each other in TG. You want them thicc-er like this?

thicc

GhostEcho:
Its how the corners just go into triangles

GoodG1448:
By looking at how wall textures link in the new map, I will guess that it will conflict with the new sprites in the new map. Closing until all the map merges and stuff are ready and set.

PR #969, opened 2019-06-11
Outcasts merge

chowdermcarthor:
## Description
Merging outcasts map and stuff from outcasts.


GhostEcho:
Is this still replacing current map, or is it a new map so we can just alternate between the two instead of reverting and remerging

PR #980, opened 2019-06-12
New test of Pahrump

chowdermcarthor:
Test of the new map lads.


chowdermcarthor:
Don't pull yet, need to fix something.

chowdermcarthor:
Alright, should be good now.

PR #982, opened 2019-06-13
Soldier's donator items.

chowdermcarthor:


pireamaineach:
Wouldn't you need to actually make the items in a DM?

chowdermcarthor:
I thought so but there's already a donor.dm with his stuff in it.

PR #997, opened 2019-06-15
Fixes an atmos breach, and a little mapping issue.

Rohesie:
This was only a warning. If our demands are not met, our next actions will be worse.


Rohesie:
Looks like #992 made a little guard house and walled it off under the door and windows. The mapper tried to fix it by deleting the wall in #994 but didn't test it. Replaced the wall with the cold, empty suck of space. And thus was the air gone.

PR #1000, opened 2019-06-16
Removes dismemberment, per Ren saying I should.

AdamElTablawy:
ren said i should. so i did.

dismemberment go bye.

also because chowder keeps accidentally depressurizing his whole map i'm givin humans the ability to resist low pressure because low-pressure environments like spce shouldn't even exist in the wastes and this should just mitigate any issues where that happens

## Changelog (neccesary)
:cl:
tweak: You can no longer dismember humans. This is temporary.
/:cl:


companoo:
if its temporary cool

Hekzder:
No longer needed.

PR #1019, opened 2019-06-17
Small Fix

Karl-Franz-I:
## Description
Made salvaged PA helmet values match the armor itself, like lots of of other armors.
Gave the NCR engineer insulated gloves so they don't kill themselves setting up additional power methods.

## Changelog (neccesary)
:cl:
add: NCR Engineer now starts with insulated gloves.
balance: Made the heavy trooper salvaged helmet have the same protection as their suit.
/:cl:


BigManClancy:
Not sure why the Engineer didn't already start with Insuls to begin with...

PR #1024, opened 2019-06-18
Mosin Buff

SeniorScore:
It's unusable trash. Changing that to usable trash. Removed autistic nature of bolting, given a small fire delay. Also one to scoped hunting rifle because that's coming anyway.

Adds a fair amount to garbage piles, mostly low-tier weapons and some ammo. Hopefully helps alleviate the issue of a lack of guns the longer a round goes on until I can figure out this blacksmithing bullshit.


SeniorScore:
NOW WITH BUFFS INCLUDED TO _ACTUAL_ TRASH

BigManClancy:
Wholeheartedly agree with this PR. Mosin is good gun, i am sad that it was so autistic before these changes.
pls approve and do not look back

Hekzder:
+1 The current Mosin causes arthritis and that is bad.

PR #1036, opened 2019-06-21
Update tips.txt

Wer6:
## Description

Murders some useless tips, adds some useful ones. 
## Motivation and Context

There was a lot of useless information (why do we need blob tips when we never blob??)
## How Has This Been Tested?

NA
## Changelog (neccesary)
:cl:
add: some of those neat roundstart tips
del: removed useless neat roundstart tips
/:cl:


Wer6:
you got it boss

PR #1043, opened 2019-06-22
Changes endround splashscreen

GoodG1448:
Changes the default /tg/ endround image for the classic "stand by" screen from Fallout

It should look like this:

like this


Hekzder:
+1 Yes.

PR #1048, opened 2019-06-23
Vault Update

brobra1:
## Description
Adds vault to the current map and re-enables vaultie roles. Vault is limited on food and uranium and will have to eventually send an expedition party to obtain more supplies and keep the vault functional. 
## Motivation and Context
Takes full use of vault assets with the redesigned vault and makes them a more traditional faction.
## How Has This Been Tested?

Vault has been compiled and tested multiple times for its various features.

## Changelog (neccesary)
:cl:
add: Redesigned vault
add: Gravity generator near Telecoms since vault can't function without it
tweak: modified overseer wardrobe
/:cl:
## Screenshots
Screenshot (72)
Screenshot (73)


brobra1:
Removed R&D from vault.

Hekzder:
This PR is incredibly fucked up and is likely the result of a really out of date master on your end, it says you're updating 29 files when in reality you should only be changing like 3-4. Your PR is pushing updates that have already happened prior, making me think you definitely have an out of date master.

Also merge conflicts, even if we wanted to merge this, we couldn't.

brobra1:
Alright, I'll look into what's causing this and provide a fix.

PR #1086, opened 2019-07-03
Makes the IC button sound like less of a dick

GhostEcho:


MrBobCool:
Brings tears to my heart.

code/modules/admin/verbs/adminhelp.dm
```diff
@@ -380,8 +380,7 @@ GLOBAL_DATUM_INIT(ahelp_tickets, /datum/admin_help_tickets, new)
 		return
 
 	var/msg = "<font color='red' size='4'><b>- AdminHelp marked as IC issue! -</b></font><br>"
-	msg += "<font color='red'><b>Losing is part of the game!</b></font><br>"
-	msg += "<font color='red'>Your character will frequently die, sometimes without even a possibility of avoiding it. Events will often be out of your control. No matter how good or prepared you are, sometimes you just lose.</font>"
+	msg += "<font color='red'>Sometimes you just have a run of bad luck, if it's not a rule break, there's not much we can do as admins</font>"
```
MarinaGryphon:
```suggestion
	msg += "<font color='red'>Sometimes you just have a run of bad luck, if it's not a rule break, there's not much we can do as admins.</font>"
```

PR #1091, opened 2019-07-08
Adds carpet and crayon production to the Biogenerator

GoodG1448:
This should give players more options to decore stuff, crayon can be grinded into powder that can be used to paint floors/items if sprayed and carpet can be used to replace the boring wood flooring. 

100% tested

Stuff

Carpet biomass requirement is low because it only makes one tile. Clicking on 10x would make 10 tiles that would cost 50 biomass.


uzername:
Roll of cloth requires 50 biomass (a single one), but carpet tile (which is more thick) requires only 5 biomass. Is it correct?

PR #1108, opened 2019-07-12
Fuel barrels now can refill welding tools

GoodG1448:
Explosive/Fuel barrels can now be used to refill welding tools. It works with all welding tools

refuel

Also, it A-logs everytime someone blows it up with a welding tool or shoots the barrel.

loggin1

100% Tested

### DIFF
--- a/code/modules/fallout/obj/structures/barrel.dm
+++ b/code/modules/fallout/obj/structures/barrel.dm
@@ -55,6 +55,46 @@
 	..() //extend the zap
 	boom()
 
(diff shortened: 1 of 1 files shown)


GhostEcho:
Do we really need admin logs for this?

GoodG1448:
I could remove the A-log part but not sure what the admin staff wants. Anyone can just drag one of those barrels and immolate themselves wordlessly in the Kebab bar tbh, atleast this will leave some logs.

PR #1114, opened 2019-07-13
Chiken Ranch Borthel

Kabramen:
## Description
Add the Chiken Ranch Borthel, northeast of kebab, with it own custum sign.

Based on the real live bothel Chi**c**ken Ranch, also located in Pahrump.
https://en.wikipedia.org/wiki/Chicken_Ranch_(Nevada)

## Motivation and Context
Since the change of map, people were requesting a new borthel, so here it is!

## How Has This Been Tested?
Tested in local server

image
image
image

## Changelog (neccesary)
-Adds Chiken Ranch Borthel
-Adds an new Sign


Kabramen:
After some testing and fixing some stuff, the borthel is finished. Also,I took the liberty and put some lights in the costumes store and removed an random door in the NCR base.
image
image
image
image

PR #1128, opened 2019-07-18
Deathclaw Cloak

GhostEcho:
changes the name of the goliath cloak into a deathclaw cloak, makes it a bit more protective as well as allowing it to store guns in it's suit slot
Can also finally craft it as well


GhostEcho:
Hekz is okay with it

PR #1142, opened 2019-07-22
meth and salt gone

AdamElTablawy:
now you can't make meth or bath salts


JohnOxford:
bad

uzername:
Why?

SpaceVampire:
This is awful even by your standards adem

AdamElTablawy:
you know full well why

JohnOxford:
very insightful
much convince

ErwinSnowden:
bath salts are perfectly balanced because their brain damage is not nullified by mannitol and you will die if you take them without moderation.

you'll die easily if you eat more than 15u of bath salts the brain damage is that bad, otherwise i'm fine with meth removal since turbo exists.

Hekzder:
This has been lingering around for a couple weeks, is probably not needed, let alone the fact that no one can really agree on whether or not this is good.

PR #1143, opened 2019-07-22
fine

AdamElTablawy:
fine ghost i'll remove just illegal modules for now


FantasticFwoosh:
In general the whole techweb i think needs a teeny tiny revision to not be dumb, the cutoff point of advanced tech & lack of "secret-tech" (which was removed from the code in favor of getting alien-tools/surgical equipment sooner) means they finish everything quickly.

I don't care for the balance of the modules, but with the danger they present in mind, on /tg/ they are disk design transferrable only to upload onto the RnD computer which involves going out to track them down (something you'd expect knights/paladins to be interested in doing with technology hoarding/confiscation)

AdamElTablawy:
that's a lot of text i didn't read for a faction i don't care about

i deleted the line of code that lets you make illegal modules

everyone else may catch fire and die

xeroxemnas:
cringe

PR #1146, opened 2019-07-22
Allows dead people to be sent out of the wasteland

JohnOxford:
## Description
This allows dead people to be sent out of the wasteland.

Fundamentally, the purpose of this is to incentivize factions collecting their corpses of important job slots, and allows the job slots to be freed up in the event a faction is beginning to make a come back, this also prevents faction positions from being removed entirely during the course of a round.

Same name rules still apply. It's logged.

## Motivation and Context
Killing a centurion or a captain basically ensures that the extent of the leadership will be run by a lower bracket in that faction or that someone will have to pick up that persons gear, which sometimes results in arguments about who deserves it and overall facilitates a unrealistic portion of command, that being that he who wears the fancy hat controls the faction.

RP wise, this basically represents that the body was taken back to -x off map faction main base- and buried, and then the faction later on (whenever a player selects the open job slot) sent a new one back onto the map.

This doesn't occur unless someone takes the time to drag the body to a matrix wall and send them back, and its not guaranteed to fill the slot anyways, as someone still has to go into it.

## Changelog (neccesary)
:cl:
tweak: Tweaks respawning bodies at matrix walls.
/:cl:

### DIFF
--- a/code/modules/fallout/turf/walls.dm
+++ b/code/modules/fallout/turf/walls.dm
@@ -180,9 +180,6 @@
 	if(!ishuman(dropping))
 		return //Only humans have job slots to be freed.
 	var/mob/living/carbon/human/departing_mob = dropping
-	if(departing_mob.stat == DEAD)
-		to_chat(user, "<span class='warning'>This one kicked the bucket. Won't be traveling anywhere.</span>")
-		return
 	if(departing_mob != user && departing_mob.client)
 		to_chat(user, "<span class='warning'>This one retains their free will. It's their choice if they want to depart or not.</span>")
 		return
@@ -205,7 +202,7 @@
 		dat += "."
 	message_admins(dat)
 	log_admin(dat)
-	departing_mob.visible_message("<span class='notice'>[departing_mob == user ? "Out of their own volition, " : "Ushered by [user], "][departing_mob] crosses the border and departs the Mojave.</span>")
+	departing_mob.visible_message("<span class='notice'>[departing_mob == user ? "Out of their own volition, " : "Ushered by [user], "][departing_mob] crosses the border and departs the Mojave Wasteland.</span>")
 	departing_mob.despawn()
 
 


SpaceVampire:
I think this is a bad idea. If you kill an important slot, then that slot should just be gone from the rest of the round. This will also break the don't respawn with the same name code, in the way it currently works.

JohnOxford:
I'm almost certain it doesn't break it, as the name is recorded

"Same name rules still apply. It's logged."

You still can't respawn with the same name even if your body is respawned, i'm pretty sure correct me if i'm wrong.

In any other situation, if you kill a important slot, and you don't kill the entire faction, another member of the faction is just going to pick up the equipment and become the slot.

Fundamentally this changes nothing except it formalizes the line of succession instead of making it arbitrary and enhances RP by incentivizing bringing back dead bodies of your faction.

The only reason you would have to dislike it is the premise that you're imagining playing a faction, neutralizing a opposing faction, then having them come back with a captain and veteran ranger for example because you failed to burn/bury/hide/destroy/take the bodies.

ma44:
If a dead leader or a restricted job class slot that's high ranking and gets looted and thrown into the matrix, that job can be respawned again *with all of the loot* once again, this can potentially lead to big amounts of high tier loot flooding the wasteland

SpaceVampire:
You are wrong, the way it works right now it is even removes your name from the crew manifest, so you can despawn and then respawn as the same person. Loot flooding is also a big issue. I shouldnt have to mutilate every single body for people to just stay dead.

GoodG1448:
Honestly, it don't see problems as long as it clears faction's cannon fodder job slots like NCR Recruit or Legion Recruit. Clearing command roles seems to be a huge issue since high tier equipment should always pass to the the second highest rank in the faction instead of spawning a new command role.

On other hand, I thought of the concept of buring people to clear job slots. If the dude is no longer optimal for revival and buried after X minutes, then the job slot gets cleared. I can imagine preachers going around the map gathering corpses and giving them proper burial in the church's graveyard.

JohnOxford:
I - that request
I
I
I
I
I
I
I - my level of coding capability

PR #1153, opened 2019-07-23
Revert "removes heal stacking and removes super stims not having an OD"

JohnOxford:
Reverts BadDeathclaw/TG-Claw#1144

no one likes this, this makes healing chemicals the best weapons in the game and prevents you from being healed if you are put into crit because people don't know which powder you used or which stim you used


Hekzder:
This will either be changed to react into something else or remain as Anacea, reverting the heal stacking change is not going to happen any time soon. Heal stacking is cancer and not a good mechanic.

PR #1155, opened 2019-07-23
Anacea metabolizes faster

AdamElTablawy:
i don't like agent's PR


SpaceVampire:
I don't think this is needed. I mean, its close enough to mine that I don't *really* care, but it just feels overly vindictive. It does also ruin Anacea if you actually want to use it as a toxin, but I will admit thats not really an issue.

JohnOxford:
i don't think you're needed

PR #1167, opened 2019-07-27
Replaces the default metal wall sprite

GoodG1448:
I did this one month ago but they said that the sprite was too triangular. Here's an alternative to that
preview

Here's how the other metal wall sprite I did looks for those interested

reference final


companoo:
ngl our present metal wall looks alright unless we really want to change it

Hekzder:
This is literally very old and no one seems to be able to come to an agreement that these sprites are better than our current ones. Talked it over with some others and this is probably just going to be closed.

PR #1169, opened 2019-07-29
Makes the 357 revolver function as intended.

pireamaineach:
## Changelog (neccesary)
:cl:
tweak: Makes the 357 revolver actually not able to load more than 1 bullet at once.
/:cl:

### DIFF
--- a/code/modules/projectiles/boxes_magazines/internal/revolver.dm
+++ b/code/modules/projectiles/boxes_magazines/internal/revolver.dm
@@ -26,6 +26,7 @@
 	ammo_type = /obj/item/ammo_casing/a357
 	caliber = "357"
 	max_ammo = 6
+	multiload = 0
 
 /obj/item/ammo_box/magazine/internal/cylinder/rev6250
 	name = "6250 10mm cylinder"


Hekzder:
Un-needed nerf for the .357 when it already is weaker than the .44 magnum and is generally fine as a T2 pistol.

pireamaineach:
Well that's a balancing issue then. Mind that these calibres were planned to be reworked when the P.R. that added these guns was made, and this is mechanically how the .357 is supposed to act. It uses a loading breech.

Hekzder:
Yes, it is a balancing issue. Making this a balance PR that is not needed in any way. We decided a long while ago that we didn't want to have .357s load one bullet at a time.

pireamaineach:
It's a colt SAA, that's how it works. The .357 isn't supposed to be better than the .44, it's supposed to be more common. How come when someone makes dogtags they get consulted on whether or not the sprites can be changed, but when I make a gun everyone can change the base of how it functions and change the sprites without me hearing anything?

Hekzder:
I don't know, it wasn't my decision that that was how dogtags were going to be handled, it was Ren's. To be fair I hardly know who you are and don't interact with you much, so keeping you in the loop of things is difficult, especially when I don't know how active you are on our Discord, where changes and things like this are discussed.

pireamaineach:
I'm Bip, mate.

Hekzder:
Ah, didn't know this was your Git user. That kind of just makes this whole situation worse then, you're a Scribe, you have access to all of our coding channels. You have full access to the discussions and knowledge regarding any of these changes, you should probably be putting that to use and keeping yourself in the loop. Regardless, this change in particular isn't really needed at all, .357 is fine as it is.

pireamaineach:
Yeah this isn't incredibly inconsistent and irritating at all. Sorry I missed 1 discussion about the .357 revolver. Guess nobody thought I should be notified as the creator, unlike the fucking dogtags.

Hekzder:
I didn't know you were the creator, I didn't know who the creator was because guns were being re-balanced and the .357 was being re-added as I was returning from my hiatus. The decision to make the .357 load like a normal revolver was one we made pretty early on while re-balancing guns and everyone was like "Yeah, that's fine, it's already worse than the .44" This is wholly un-needed. 

Regarding the dogtags, it's something you can take up with Ren and/or Senior. I don't really care if dogtags look different or not.

pireamaineach:
So why wouldn't you just buff the calibre instead of making the gun function in a way that it simply doesn't? It's a Single Action Army, it shouldn't be able to load multiple rounds at a time. It can do that shit from loading tubes as is!

pireamaineach:
Also if it's okay for you to miss something while you're on Hiatus, why isn't it okay for me to miss something? Why is my fault that nobody talked to me while I was talking a break from B.D., but you're to be forgiven because you were on hiatus when the gun was added?

Hekzder:
I'm not entirely saying it's your fault, but you aren't very active in any of our Scribe/coding discussions, well, at all, thus you're probably going to be left out of such discussions as a result and therefore be left out of the loop. I'm not particularly asking for forgiveness or trying to paint it as "okay" that I went on hiatus, I'm stating that I did not know you were the creator and the reasoning why. 

Nobody talked to me directly regarding changes during my hiatus either, I didn't have an expectation that they would. All I'm saying is that if you do go on a break and don't keep up with what is going on, you shouldn't particularly be surprised that things have transpired in your absence, potentially ones that you didn't want to happen.

pireamaineach:
Okay, so why close the P.R. instead of having a discussion about it? Why completely change the gun from what I originally wanted and not even be willing to change it back?

Hekzder:
Because, as I already stated, this is an un-needed QoL and even balance change. Loading guns one bullet at a time is inherently more annoying, let alone objectively inferior to a speed loader. This would be a negative change both for the gun itself and the people using it.

pireamaineach:
It's not supposed to be a better gun, it's supposed to be a more common gun. A gun being worse than other guns isn't intrinsically a bad thing. Why don't we have full auto mosins? Them being bolt action is worse for the players using them!

PR #1181, opened 2019-08-03
Techwebs localhost fix

Arkatos1:
## Description

This PR fixes a case where  players could not play with techwebs when locally hosting, because they did not properly initialize. This fixes that along with the corresponding runtime.

## Motivation and Context

Less runtimes, all players can play with techwebs on their localhost.

## How Has This Been Tested?

Locally without any issue.

## Changelog (neccesary)
:cl: Arkatos
fix: Fixed a case where techwebs nodes did not properly initialize.
/:cl:


GhostEcho:
Bless you

PR #1202, opened 2019-08-08
Muzzle speed overhaul

Tytynguyen:
## Description
Adds muzzle speed to guns, allowing for varied projectile speeds for each weapon. All guns start with the base projectile speed, though it can be changed by setting the muzzle_speed_modifier value for a weapon or weapon group. This update also adds some initial values to muzzle speeds for various guns. The speed of projectiles from gun groups is generally Snipers>Rifles>Pistols/Submachines where pistols and submachines are the standard speed. Lasers, shotguns, and many other guns do not have custom muzzle speeds in this update, but they may be added for balancing in future updates.

## Motivation and Context
Adds a deeper layer to the gameplay, buffs snipers and long-range scope-based shooting. Allows for more customization in future updates.

## How Has This Been Tested?
Tested weapons ingame to determine good speeds for each weapon.
Manually checked that variables were added correctly to weapons.

## Changelog (neccesary)
:cl:
add: Added the basic muzzle speed functionality
add: Custom muzzle speeds for a handful of weapons
/:cl:


RenGusta:
I really would have preferred an addition of the variable, and leaving the Nitty gritty balancing for the remainder of the Dev Team. If you could do that for me, that'd be good.

Along that line, I looked at your code. Is the 1 in deciseconds? A little documentation saves a lot of trouble.

As I understand, base speed is 0.8 deci for bullets, so increasing it by 25% makes sense for this change, but it's a pretty substantial balance shake up in the meantime.

Tytynguyen:
> I really would have preferred an addition of the variable, and leaving the Nitty gritty balancing for the remainder of the Dev Team. If you could do that for me, that'd be good.
> 
> Along that line, I looked at your code. Is the 1 in deciseconds? A little documentation saves a lot of trouble.
> 
> As I understand, base speed is 0.8 deci for bullets, so increasing it by 25% makes sense for this change, but it's a pretty substantial balance shake up in the meantime.

I thought that might be the case. I have reverted those initial values, so everything should be at default now. I also realized that the multiplier in its previous implementation may have caused a bit of confusion to use, since it was literally multiplying the base speed (1/deciseconds). Therefore, I changed the implementation so the value is truly a multiplier for the projectile velocity. A multiplier of 2 will double how fast a projectile flies, instead of making it slower.

RenGusta:
Just to confirm, the speed 1 is at 0.8 Deci per tile, and not 1 Deci per tile, making the modifier based around the 0.8 Deci. Yes?

Tytynguyen:
> Just to confirm, the speed 1 is at 0.8 Deci per tile, and not 1 Deci per tile, making the modifier based around the 0.8 Deci. Yes?

That is correct. The modifier is based upon the 0.8 deci base speed.

RenGusta:
If you can comment that into the code near the muzzle_speed_modifier to clear it up for any future readers, that'd be excellent, and it should be clear to merge.

Tytynguyen:
I've found a much cleaner and nicer implementation of projectile speed. Instead of a multiplier, why don't we just set the speed directly? Huh!

Anyways, I'm closing this PR and opening a new one with the other implementation which is much easier to read/use

PR #1203, opened 2019-08-09
Adds a new trait, 'Power Pitcher', for thrown weapons

CameronWoof:
## Description
This PR adds a new trait called Power Pitcher that increases the damage dealt by throwing weapons by 5 points. It costs one trait point (half as much as Big Leagues, which applies to all melee weapons).

## Motivation and Context
Recruit Legionnaire is a fun playstyle, but none of the existing traits synergize particularly well with it. With the addition of this trait, career recruits can stick to their talents and gain an edge for doing so.

## How Has This Been Tested?
Spawned in and chucked spears at human and simplemobs to verify that the damage boost works when you have the trait and *not* when you don't. No runtimes.

## Changelog (neccesary)
:cl:
add: A new trait called 'Power Pitcher' has been added for fans of throwing weapons.
/:cl:


Hekzder:
Not really sure how useful or needed this trait is. Also generally not a fan of adding or messing with the current quirk system as it's genuinely awful and just needs a full overhaul. It's hard to fix or add to a system that is inherently garbage.

PR #1207, opened 2019-08-09
Adds projectile speed implementation for guns (Round Two)

Tytynguyen:
## Description
Retry at adding custom projectile speed to guns, allowing for varied projectile speeds for each weapon. All guns start with the base projectile speed of 0.8 deciseconds per tile, though it can be changed by setting the projectile_speed value for a weapon or weapon group. Setting this value for each weapon or weapon group will directly set the speed in the deciseconds per tile units, rather than my workaround in the previous implementation with a modifier. Hopefully this makes the projectile speed MUCH easier to use and understand in context.

## Motivation and Context
Adds a deeper layer to the gameplay. Allows for more customization in future updates.

## How Has This Been Tested?
Manually checked that variables were added correctly to weapons, and tested different speeds for various guns

## Changelog (neccesary)
:cl:
add: Added the basic projectile speed functionality
/:cl:


RenGusta:
I am proud of you son

PR #1208, opened 2019-08-09
[MAP CHANGE] Adds support for Legion radios

CameronWoof:
## Description
This adds the Legion radio frequency (:e) and the telecoms machinery required to support it. This does NOT make it available for use by players, except by admin-spawned encryption keys should they wish to test it out.

## Motivation and Context
Legion having their own channel would be nice, but there's no quick and easy way to balance it. This will add the basic support immediately, and the actual implementation can be decided on for a later PR.

## How Has This Been Tested?
Spawned in, tested the channel. It shows up and is the correct color.

## Changelog (neccesary)
:cl:
add: Adds support for a new Legion radio channel (not available to players)
/:cl:


code/game/objects/items/devices/radio/encryptionkey.dm
```diff
@@ -168,6 +168,13 @@
 	icon_state = "cypherkey"
 	channels = list("Kebab" = 1)
 
+/obj/item/encryptionkey/headset_legion
+	name = "Legion radio encryption key"
+	desc = "An encryption key for a radio headset.  To access the Legion channel, use :v."
```
Arkatos1:
```suggestion
	desc = "An encryption key for a radio headset.  To access the Legion channel, use :e."
```

CameronWoof: Good catch. This is fixed now.

PR #1211, opened 2019-08-09
[MAP CHANGE] Adds the compost bin and low-tech seed extractor

CameronWoof:
## Description
Adds two new objects and one reagent, focused on making legionnaire farming methods less reliant on stolen technology.

The compost bin is a structure that spawns in the legion camp and accepts seeds and produce, which it then turns into *compost*.

Compost is a reagent that is high in nutrition and helps ward off pests (which the legion have no natural way of doing).

The seed grinder is a low-tech machine that turns flowers into seeds. It's less efficient than the real thing, and doesn't have all of its features. 

## Motivation and Context
At the moment, legion farming is virtually impossible to do without either using advanced technology (which legionnaires shouldn't be doing) or abusing sand to make infinite, fully-watered and nutritious soil plots. With these objects, it's now possible to go without either.

## How Has This Been Tested?
Spawned in-game, ensured that all the basic interactions work with both machines, that you can unwrench and move the seed grinder, and that the compost bin behaves properly when full. It's looking pret-ty good.

image

## Changelog (neccesary)
:cl:
add: Added the compost bin, a new structure for generating nutriment for plants.
add: Added the seed grinder, which is a low-tech seed extraction device.
tweak: The legion camp now spawns with plant bags instead of seed extracting bags, in light of their new machine.
/:cl:


Hekzder:
So, while I like the sprites and overall intent behind this, I feel as though it will be rendered completely useless. The Legion isn't really _prevented_ from still going and collecting a technological seed extractor, nutrimax, seed vendor, etc. And with these both being inferior to their technological counter parts, they have no reason not to.

The only solution I can think of to this is really just to prevent Legion from using said seed extractors, nutrimaxs, seed vendors, etc. either through rule enforcement or perhaps specific code to do so, a solution I really don't think is either needed or is a solution that I'm all too happy with.

CameronWoof:
If we're playing by the rules, we shouldn't be using this kind of technology--I figure it was being overlooked simply because there's no good alternative to it. If these options are implemented, admins have good reason to say "No, you don't need those, put them back."

Hekzder:
I can probably speak with some admins about this and see what kind of conclusion we can come to, because this is in no way purely a code issue and definitely does boil down to an issue of "What do we allow the Legion to use and not use?" So I'll try and get back to you on that once we do have some discussion.

RenGusta:
Speaking happened.

PR #1212, opened 2019-08-09
Reworks healing item conflicts

CameronWoof:
## Description
This PR removes conflict reactions in favor of a new method that doesn't allow chem stacking, but doesn't punish the attempt. This PR establishes the following hierarchy, from best to worst:

1. Super Stimpak
2. Healing Poultice
3. Stimpak
4. Healing Powder

Chems higher in the hierarchy will not heal (but will still deplete) if a chem below it in the hierarchy is present. This means that *only* the weakest chem heals, but reagents aren't "deleted" in the case they conflict.

## Motivation and Context
Anacea leads to unfortunate situations where someone who has already used a stimpak/healing powder on themselves is then treated by someone who cannot know they've already done so, and the result is that two healing items are destroyed and the wounded simply doesn't heal. 

This approach remedies the previous issue (chem stacking) by instead having the weakest chem simply override the stronger ones. The player being treated always receives healing, but it's not always the best possible outcome, and there's no situation where counter-intuitively, medicine is applied and *nothing happens*.

## How Has This Been Tested?
Spawned in and verified that only the strongest chem in the mob's system affects its healing rate, but they all still deplete as normal.

## Changelog (neccesary)
:cl:
balance: Removed the anacea reactions from the code. Instead, only the weakest healing item will apply its effect. Super Stimpak > Healing Poultice > Stimpak > Healing Powder.
/:cl:


Hekzder:
Not really a fan of this, heal stacking should still be some level of punishing, chems shouldn't persist despite stacking, and anacea in it's nerfed state is fine. Don't really think this is needed and I don't think there are many complaints about the current existence of anacea as the reaction, to my knowledge.

CameronWoof:
I think the problem with anacea is that it's not well-targeted. People who want to abuse chem stacking need only be told "hey, that doesn't work" and they won't waste their time. Anacea *mostly* affects people who aren't trying to stack chems, but do so unknowingly by accident. After a fight with the legion, I can't tell which of my downed legionnaires pre-powdered themselves (in which case the poultice I have will only hurt them) and which ones haven't (and are therefore on their oxygen spiral to death). 

This effectively nullifies the stacking while also making sure that applying medicine is *always* helpful, since players usually won't have any way to read a critical person's reagents.

Hekzder:
While I get the point that nullifying chem stacking is enough to get people to not want to do it, Anacea is still fine for the reaction. If you don't know what kind of medication someone is on, then don't give them medication, that is proper medical practice in the real world due to potential complications with giving someone a conflicting medication, IE exactly what we're looking at now. 

The current iteration of Anacea is not incredibly harmful due to not doing toxin damage and not lasting as long and I generally think it's fine as is, heal stacking should be punishing beyond just "Oh, my chem got nullified" especially when your solution, with the hierarchy, would still allow one of the healing medicines to persist, to my understanding.

CameronWoof:
The chems still process out (so if you took ten units of each at the same time, they'd all be gone from your system at roughly the same time), they just don't heal you. The problem isn't that you *get* anacea, it's that most of the healing chems in you are *removed* to perform the reaction--and *then* the anacea increases the depletion rate of what's left. 

There's simply no way for me to tell, outside of trying to secure a medical scanner, if someone dying in front of me is getting better or getting worse. I think the best thing that the game can make you want to do at that point is help them, and anacea does the opposite--it forces you to hesitate and waste time because, who knows, they *might* be getting better on their own. 

Anacea was designed to stunt powergaming, but I don't think it hurts powergamers anywhere near as much as it hurts casual players, players in need of help, and players who want to help others.

PR #1225, opened 2019-08-11
Fixed using super stims to craft super stims

bardinatavern:
how is this even an issue. I can't even begin to understand how someone fucked up this badly.


CameronWoof:
Map changes will also need to be made to fix the paths of the super stims on-map, correct?

Arkatos1:
Cameron is right, there are about 10 instances of super stimpack placed on the map, they also have to be properly renamed before this is merged.

PR #1238, opened 2019-08-15
Incombustible lemons 1

Thescri:
## Description

Commits are a mess again because of that stupid botched xeno change mess, but this comments out the mutatelist line that allows lemon trees to be mutated into combustible lemon trees. Pouring mutagens like FEV on them can still cause pest and weed mutations, I assume, but no amount of FEV will produce a combustible lemon now. Left the seeds and trees, as well as the actual produce, intact in case someone wants to spawn them for somebody, but I'm pretty sure that the mutation is impossible. 

## Motivation and Context

Combustible lemons cause chaotic combat logs and confusing situations fairly often, so it's better if they just don't exist. They would also be too easy to create compared to grenades made through any kind of potential grenade crafting system.

## How Has This Been Tested?

Tested locally with no issues and no mutations even after pouring about 200u of unstable FEV solution onto a lemon tree, obtained both through chemistry and grinding up ghoul meat.

## Changelog (neccesary)
:cl:
del: Commented out the mutatelist line under lemons
tweak: New comment stuff in citrus.dm, mentioned just to be safe
/:cl:


GhostEcho:
Explosion rebalance in the works, rejected after consultation with the balance lead.

PR #1250, opened 2019-08-18
Weather hotfix

Arkatos1:
## Description

This PR hotfixes weather, which had a case of not showing their effects in correct areas due to my mistake in the last PR. (sorry)

Also contains a fix for a soil exploit, where one could constantly get fully watered nutrient soil by digging and then crafting the soil again, bypassing any need for water, nutrient or other tools for growing the plants.

## Motivation and Context

Workable weather.

## How Has This Been Tested?

Locally all worked well.

## Changelog (neccesary)
:cl: Arkatos
fix: Weather hotfix.
/:cl:

### DIFF
--- a/code/datums/weather/weather.dm
+++ b/code/datums/weather/weather.dm
@@ -17,7 +17,7 @@
 	var/weather_overlay
 	var/weather_color = null
 
-	var/end_message = "<span class='danger'>The wind relents its assault.</span>" //Displayed once the wather is over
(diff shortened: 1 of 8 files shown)


CameronWoof:
I'm not currently in favor of fixing the soil "exploit".  It's currently the only way to reliably start a farm from the ground-up if you don't have ready access to a biogenerator or vending machine.

GhostEcho:
The fix is fine, you can mix water and milk for good fertilizer and do composting now

PR #1256, opened 2019-08-19
Economy status: online

Arkatos1:
## Description

This PR furthers the economy system brought in the last PR, feedback from players was also considered:

1) Trash pile changes - Trash piles no longer replenishes themselves indefinitely. This caused over-abundance of items, making them small value. When trash stack is depleted of its contents, it will disappear, which also fixes an issue where players could farm them when housing near them, or it also caused issue where it blocked the tile it was on, preventing players from building in the area it was located. You can also no longer insert any item into the trash pile, which could cause them to disappear under the old system.

2) Trade Protectron changes - Trade Protectron will no longer aggro hostile mobs, his speak chance and even speak lines were updated, plus minor tweaks.

3) Wasteland Vending Machines(WVM) changes - Updated general descriptions, fixed a bug regarding rounding values and a little cleanup. Also fixed a bug where certain items could become unavailable and where lights could go to emergency mode in buildings where they are not supposed to. Weapons WNM now also accepts guns with their magazines loaded in them.

4) Added 5 new types of WNMs, as was planned. Currently, we had only medical type. This update adds Weapons, Ammo, Clothing, General and Special types of WNMs with their own sprite and assortment of items.

5) Added 2 more trading protectron shops in the Wasteland, each having 2 kinds of vendors, which brings us to 3 shops total, which covers all types of WNMs.

After this update, the main remaining thing that will be left is to properly balance, and consider which items WNMs should sell, but functionality wise, nothing much else is there to add.

Amen.

## Motivation and Context

Bringing economy aspect to the game.

## How Has This Been Tested?

Locally tested without an issue.

## Changelog (neccesary)
:cl: Arkatos
add: Economy systems fully activated. New vendors and Trading Protectrons were added.
/:cl:


SpaceVampire:
Can you please remove the emotes? I find them to be kind of irritating and repetitive. Its just spam when I am observing.

Arkatos1:
@SpaceVampire You are right, emotes really do not bring anything interesting. I have removed them, so nly his speak lines remain, albeit with a reduced chance.

Issue #1166, opened 2019-07-27
High-Lethality is problematic


UndeadNappist:
Removing easy stuns and incapacitates mean everyone will resort to shooting each other.

This isn't necessarily all bad, as many people will use the cheap stuns into an simply methodology of murder, but having the ability to incapacitate/kidnap each other with relative ease prevents characters from being removed from the round if someone plans something.

Increasing the ability to act while in crit further exasperates this problem and makes it more likely for people to fight to the death.

Possible solution:
Increase amount of health available while unconscious
Make it so you heal while unconscious (at least, while sated, or perhaps other mechanics) and remove the O2 loss while unconscious (leaving bloodloss as a possible death)
Remove people's ability to tell whether you're dead or just unconscious, including the deathgasp thing unless they take the time to check your pulse ; deathgasp could be left in when leaving your last whisper or succumbing deliberately.

RenGusta:
You know I actually like this, Hopefully someone starts pulling some ideas from this.

PR #1265, opened 2019-08-22
Decreases oxyloss, increases crit amount

UndeadNappist:
Does what it says on the tin.

#1166


CameronWoof:
Increasing the damage needed to kill someone by 50% is a large balance change that shouldn't be part of this PR. Take just the reduced oxy loss for now and start doing other stuff after.

UndeadNappist:
This PR probably doesn't go far enough to do all the things I want it to do, and likely isn't sufficient enough in all the alterations.

"Because it is a large balance change" is an insufficient argument to deny something.

cross-ref;
https://github.com/BadDeathclaw/TG-Claw/pull/283
https://github.com/BadDeathclaw/TG-Claw/pull/612

CameronWoof:
I'm not saying a PR can't contain a large balance change, but this one contains two. You've tripled the time it takes to die from being in critical condition and THEN doubled that in the same PR. Take the first change on its own and see if that's enough. At that rate, *any* healing should be enough to reverse the brute/burn faster than the oxyloss mounts.

UndeadNappist:
No, I've cut oxy damage from not having sufficient oxygen by a third, and doubled the amount of damage you have to take in crit before dying.

Bloodloss is still an ongoing and unchanged issue.

It's not the same as "tripled the amount of time it takes to die in crit."

CameronWoof:
You will absolutely die of crit oxyloss before you bleed out, even with this change. "It takes six times longer to die" is very accurate.

UndeadNappist:
https://github.com/BadDeathclaw/TG-Claw/blob/d6ffeb207849abcc3d993d4326ee0cef6c9c1a73/code/modules/mob/living/blood.dm#L54

Bloodloss straight up adds oxyloss damage, which is not so much of a bad thing when you aren't already suffocating. This is likely why people are dying so fast in crit, and why it seems to scale up so quickly.

BLOOD_VOLUME_NORMAL is 560
BLOOD_VOLUME_BAD is 224, which is where you start taking oxyloss
BLOOD_VOLUME_SURVIVE is 122, which is where you start having chances to have heart failure and insta-die.

bleed_rate varies
https://github.com/BadDeathclaw/TG-Claw/blob/d6ffeb207849abcc3d993d4326ee0cef6c9c1a73/code/modules/mob/living/blood.dm#L76
but essentially the number of times you are hit directly impacts how fast you bleed out and how fast you take oxy damage.

It will not take six times longer to die due to this game's method of scaling how fast you die in crit upwards.

CameronWoof:
In current code, dying by bleeding out (drop into soft crit after being shot three times with a .308 rangemaster) takes between four and five minutes, as I just tested. That speed is hastened by the three oxyloss per tick of falling into hardcrit. 

Keep in mind, I died due to oxyloss, not heart failure, at over 400u of blood.

This PR, in its current state, will make that slow, inexorable slog to death take even longer. This *will not be fun*. Put the death mark back at -100 and this PR will be much less annoying to deal with.

UndeadNappist:
You can instantly die whenever you want from typing succumb. There is no slog. It's just a faint sliver of hope.

This is to make kidnapping people easier and keeping certain niche storylines open, as I said in my open issue. It is unrelated to people dedicated to killing each other.

GhostEcho:
You can also ghost as well or *deathgasp to succumb(if youre in deep crit), so the slog of dying isnt relevant, but im curious on Rens thoughts

RenGusta:
It was discussed in VC which is why he actually made the PR, dingus

CameronWoof:
This was the last thing mentioned, I said "I don't like the sound of that," and then no further conversation was made. I'm offering my feedback.

AdamElTablawy:
Good thing nobody cares since your feedback is without value

PR #1275, opened 2019-08-25
Cleaner Crafting Recipe

TMR-SS13:
## Description
Adds a method to craft Abraxo Cleaner, using the abraxo junk item, two sheets of plastic, 50 units of water, and a welding tool.

## Motivation and Context
The wasteland tends to be covered in blood most rounds, so I figured I'd add a way to introduce cleaner into the round that didn't rely on chem machines, which tend to be destroyed. This also adds some more usage for plastic, which is pretty unused most rounds.

## How Has This Been Tested?
I tested it on a private server after checking the actual code for errors. No issues were apparent, and the process worked fine.

## Changelog (neccesary)
add: Added a new crafting recipe for Abraxo Cleaner.


RenGusta:
This is fine. Assuming checks pass.

PR #1282, opened 2019-08-27
Removal of CQC from the Centurion job

DanielMcFluffy:
## Description
The Old Manual item will not be in their backpack to those that spawned as the Centurion.

## Motivation and Context
Until martial arts are reworked, it will be removed due to balance.
## How Has This Been Tested?
I deleted a line.

## Changelog (neccesary)
:CQC removal.


RenGusta:
this is fine

PR #1291, opened 2019-08-30
no more meth or eph

AdamElTablawy:
meth and eph are gone


AdamElTablawy:
yeah immie a second while i fix this

PR #1297, opened 2019-08-31
Gives Legion armors and helmets radiation resistance.

GrimmestDays:
NCR armors will likely require a later PR for tweaking. Ranger's in particular have far too high of radiation resistance. But that's a PR for later.


RenGusta:
Solved by another recent merge, #1420

PR #1317, opened 2019-09-04
MedX changes, associated minor chem nerfs

AdamElTablawy:
Ren's a retard so I just gave up and now med-x gives you 20% damage resistance to brute and burn while it's in your system it also does tiny amounts of toxin and it's addictive.

Legion medX has no changes.

Synthflesh now converts 66% of brute and burn to toxin damage if the mob it's used on is alive, to prevent synthflesh beakers being literally the absolute best healing chemical in the game.

Stypic and silver sulfadine no longer heal on touch at all, for much the same reason, but still heal over time.

Morphine has lost noslowdown for much the same reason-irrelevant since nothing uses it, but still. 

Ephedrine no longer speeds you up.
## Changelog (neccesary)
:cl:
del: Removed the effects of morphine and ephedrine
del: Removed stypic powder and silver sulfadine healing on touch (still heals over time)
tweak: Removed Synthflesh not converting 66% brute/burn into toxin when used on a LIVING person. Works the same as always for corpses. Still heals on touch.
balance: Med-X no longer removes slowdown. Instead, it increases damage resistance.
balance: Removed med-X not being instantly addictive, as it is now addictive no matter how much you take. It is also somewhat toxic.
/:cl:

### DIFF
--- a/code/modules/crafting/recipes.dm
+++ b/code/modules/crafting/recipes.dm
@@ -741,6 +741,14 @@
 	tools = list(/obj/item/lighter, /obj/item/reagent_containers/glass/beaker)
 	category = CAT_DRUGS
 
+/datum/crafting_recipe/natural_painkiller
+	name = "Natural Painkiller"
+	result = /obj/item/reagent_containers/pill/patch/naturalpainkiller
+	time = 30
+	reqs = list(/obj/item/reagent_containers/food/snacks/grown/feracactus = 2, /obj/structure/flora/wasteplant/wild_agave = 1, /obj/item/reagent_containers/food/snacks/grown/fungus = 1)
+	tools = list(TOOL_FORGE)
+	category = CAT_DRUGS
+
 
 /datum/crafting_recipe/rags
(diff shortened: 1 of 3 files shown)


Isy232:
One thing to note is that if you were to have 60 brute damage and then used a synthflesh patch or beaker you would still have 40 toxin damage. This would make synthflesh pretty much useless except for getting corpses back to defib.

If you can code it, I'd think a better solution to make the healing only 33% or 20% effective if applied by beaker so patches of these chems are useful, considering patches take time to apply.

AdamElTablawy:
@Isy232 It's meant to be useless except for defibbing corpses.

AdamElTablawy:
don't bother downvoting i don't care about feedback

AdamElTablawy:
dawn of the final day i got feedback from ren and i'll fix the conflicts after this last day with that feedback unless someone has anything else they want to say that i'll ignore

AdamElTablawy:
travis is a shit one sec

PR #1330, opened 2019-09-11
Fix Deathclaws Clipping Through Walls / Hitboxes

VendettaSpooky:
Deathclaws seem to clip through walls - and especially in caves they become invisible at times due to this being a fact. If not invisible - you can barely see their feet peaking out of a wall.

Additionally - it seems as though to attack a deathclaw you have to aim at the lower left hand corner of the sprite - otherwise the projectile will simply phase through it and miss.


RenGusta:
Has a conflict in pits.dm, if you can fix that real quick I'll get it merged.

PR #1355, opened 2019-09-18
i finally got the med-x thing to work

AdamElTablawy:
ignore the ugly commits

med-x now increases damage resistance when in your system but comes with increasingly dangerous side effects once it leaves your system the longer it's in you

## Changelog (necessary)
:cl:
balance: Med-X now increases your damage resistance by 20 to 25%, compounding with armor, while it is in your system, but no longer removes slowdown.
balance: Med-X is instantly addictive on use. On the bright side, the longer it is in your system, the more fun you will have! Keep med-X in your system as long as possible to have the most fun.
/:cl:


AdamElTablawy:
IT WORKS

PR #1361, opened 2019-09-21
fixed med-x so that ren's hemorrhoids stop acting up

AdamElTablawy:
fix


RenGusta:
nice and vague, great to track how things moved, mhmm, very helpful

PR #1382, opened 2019-09-25
Military boots

Thescri:
## Description

Finally found both types of faction shoes and added the noslip version to the Legion's kits to match the NCR's.
## Motivation and Context

The NCR currently has noslip shoes/f13/military/ncr version shoes, while the Legion's boots were all shoes/legionmetal without the inheritances from shoes/f13/military, as Crusader White had pointed out.
## How Has This Been Tested?

I've spawned both types before so there shouldn't be any issues, but it hasn't been particularly tested.
## Changelog (necessary)
:cl:
tweak: military boots for Legion instead of plain shoes that simply look like boots
balance: noslips for both major factions now
/:cl:


SpaceVampire:
Better idea, just remove no slips from NCR because that is way more funny.

PR #1407, opened 2019-10-02
bombs now yeet you

AdamElTablawy:
and no longer knock you out

this should make bombs more fair and balanced tm


RenGusta:
didn't pass travis, try again

PR #1431, opened 2019-10-08
Change SS13 radsuit for NV styled one

DIamonnd:
Basically waht the title says, I actually forgot how to do PRs so this may be wrong.


DIamonnd:
Yup this is wrong, I hate git.

PR #1439, opened 2019-10-09
Belt crafting

Technobug14:
## Description
Adds a new subcategory to crafting, as well as crafting recipes for three belt items.

## Motivation and Context
There's a lack of belts sometimes, and the only abundance of them is in faction bases. I don't like wasters using faction belts, so I'm adding this. I also hate the Brotherhood of Steel so they no longer get chest rigs exclusively.

## How Has This Been Tested?
Tested with zero errors. Vars are correct and it should all work smoothly.

## Changelog (necessary)
:cl:
Adds belt crafting
Adds recipe for shoulder holster, chest rig, belt. More soon, mayhaps.
Adds a new subcategory for new belt recipes. Should be easy to tack on new recipes, now.
:/cl:


Technobug14:
Found where bandoliers exist. Doing a quick edit to include them in the belt-crafting section. I hate whoever designed that.

PR #1442, opened 2019-10-11
obliterates BoS R&D at long last

AdamElTablawy:
most of the machines they build with R&D are otherwise available so they aren't absolutely starved. that means they start with a plant DNA modifier, DNA vault, chem dispenser, etc.

other things of note:

their armory now also has a civilian clothes vendor and a clothesmate so that BoS can disguise themselves in a way that isn't deathclaw cloak-tactical boots-tactical rig-laser pistol-roving trader armor

it also has one extra tribeam, some extra sets of nightvision, and some extra ammo, to compensate for them being unable to print additional guns and ammo from the lathe

they also have a self destruct bomb in the elder's office. it can't be moved and setting it off should level a fair amount of the base.

they also have a miniature virology lab, to give scribes something else to be autistic with.

## Changelog (necessary)
:cl:
add: Added a bomb to the BoS base so they can all kill themselves out of boredom
del: removed BoS R&D
/:cl:


boodaliboo:
HI ADAM

PR #1461, opened 2019-10-15
caps raiders

AdamElTablawy:
at 20.

20's a nice even number, half of the previous 40, should reign in raider rdm a little


Difrit:
Kneejerk change that serves no purpose, raider RDM is already handled by staff.

AdamElTablawy:
i don't think chaznoodles understands that escalation is barebones

Anton-Kr:
make it scale with amount of players t b h

PR #1462, opened 2019-10-16
NCR Base Wall Rebalance

SpaceVampire:
## Description
Removes unbreakable walls. Replaces them with multiple layers of strong yet breakable walls, based on the Kebab design. I also swapped the NCR armory location with the NCR barracks to keep people swift on there feet, and because I plan to rework the base more after this.


AdamElTablawy:
what about legion

SpaceVampire:
I'll get to them soon. For the moment, they already have two ways into the base, the main way, and the side mining tunnel.

PR #1468, opened 2019-10-21
no more assault carbine loadout

AdamElTablawy:
they're just better heavy troopers with it and there's no fixing a burstfire gun because the entire issue is that it's a multi-shot weapon


Knightfall5:
Removing features is bad when all you need to do is tweak a few numbers

PR #1469, opened 2019-10-21
Nerfs assault carbine

Knightfall5:
Increases burst delay from 1.4 to 2, and reduces damage from 25 to 20. Balance is better than removal.


AdamElTablawy:
fixes nothing

Dioclex:
did anyone except Legion mains ask for this?

Knightfall5:
I'm an NCR main, I think most people agree it's a bit too strong. And this is better than the outright removal option

AdamElTablawy:
@Dioclex i'm not a legion main and i want it

i don't really care what factions think anyway

i still insist this doesn't solve the problem though, the problem isn't just the damage the gun does

RenGusta:
Undergoing tweaking in a larger balance pass.

PR #1482, opened 2019-10-25
Insect armor (recreated to avoid conflicts)

ChronicPwnage:
## Description

Adds insect chitin and a set of craftable insect chitin armor, with the aim to expand wasteland crafting, medicine, and the effects of different types of armor in the future.
## Motivation and Context

It's cool, and I will expand wasteland crafting in this way in the future.
## How Has This Been Tested?

ran it, it ran
## Changelog (necessary)
:cl:
add: Insect chitin, which can be grinded for sodium and chlorine (hint hint, add water and ash from a fire to make your own wasteland charcoal!)
add: Insect Chitin Armor! It's got decent protection, and will be expanded on in the future.
/:cl:


RenGusta:
You do know this bug suit is currently tougher than combat armor, right? Can you drop the stats to 50/30/20 and give it some acid resist?

PR #1490, opened 2019-10-30
someone made me mad in discord so i'm capping raiders at 10 to prove a point

AdamElTablawy:


RenGusta:
Staff will manually test reduction of raiders in higher pop rounds to emulate this function.

boodaliboo:
HI ADAM

PR #1505, opened 2019-12-12
Implements drag delay

Rohesie:
Moving while pulling, pushing and moving through clicking while pulling all cause a movement delay on the user.

Lockers have the highest delay, 0.8 seconds.
Undefined structures and machines the second, 0.6. This might need tweaking for things that need lower delays.
Small items, rolling chairs, bodybags and people have the lowest delay, 0.1. It's noticeable and someone giving chase will likely catch up to them.
Rolling beds have no wield delay.
Other undefined things, including large weapons, have a drag delay of 0.3, which is very non-viable in combat.

Bodybags only accept mobs for contents now, to avoid using them as a low-slowdown object carrier.

### DIFF
--- a/code/_onclick/other_mobs.dm
+++ b/code/_onclick/other_mobs.dm
@@ -89,8 +89,8 @@
 	for(var/datum/mutation/human/HM in dna.mutations)
 		HM.on_ranged_attack(src, A, mouseparams)
 
-	if(isturf(A) && get_dist(src,A) <= 1)
-		src.Move_Pulled(A)
+	if(pulling && isturf(A) && get_dist(src, A) <= 1)
+		Move_Pulled(A)
 
 /*
 	Animals & All Unspecified
--- a/code/game/atoms_movable.dm
+++ b/code/game/atoms_movable.dm
@@ -2,6 +2,7 @@
 	layer = OBJ_LAYER
 	var/last_move = null
 	var/anchored = FALSE
+	var/drag_delay = 0.3 SECONDS
 	var/datum/thrownthing/throwing = null
 	var/throw_speed = 2 //How many tiles to move per ds when being thrown. Float values are fully supported
 	var/throw_range = 7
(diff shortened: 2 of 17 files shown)


Zebraman:
I tested it, this is pure cancer, even small items like broc flowers give slow down while pulled. Just don't.

Anton-Kr:
This server is already dead. Please don't make it more dead.

CptPatriot:
This is an abysmal idea and all it does it make the game 10x longer than it should be for no reason. DR did this for everything and it's just the absolute worst.

Rohesie:
> I tested it, this is pure cancer, even small items like broc flowers give slow down while pulled. Just don't.

This is intended. If you are unable to pick up a broc flower it means all of your storage containers are full, hands included, and thus you are pulling it with the butt.
Seems like a minor issue, considering you can even carry three backpacks on yourself (one on back, one on each hand) at no slowdown cost.

> Pulling a machine or vendor slow down under walking speed. If you have to move two of them at the same time the slow down stacks and you can barely move. This make the creation of player made settlement take forever, especially if you choose an exotic location to build it.

This is intended. Machines and vendors are supposed not to be trivial to move or steal. They are made able to be constructed and deconstructed for the most part precisely for that.
Is there a particular kind of machine you feel should be faster to move?

> Also, bringing the chem machines back to the NCR base is painfully slow.

This is intended. It has been a common complaint how fast the NCR yoinks the chem machines every round. There's a reason they don't start with one in their base anymore.

> Also, pulling animals, like brahmins and chickens slow you more than runing around with a gun unholstered.

This is true in the current implementation, an unholstered weapon will add 0.06 seconds move delay, while mobs add 0.1. If you want specific values changed, run them through Ren and I'll do them.

> If the intent of this PR is to prevent peoples from hoarding loot just apply it to container like crates, fridge and lockers.

Those are the ones with higher drag delay right now.

PR #1517, opened 2021-02-10
Removes the last of the directly mentioned lentext

FantasticFwoosh:
Weeded out via word sensitive search, the code compiles successfully with no errors checked after making my changes.

If there's anything left behind, it'll be hidden within blocks of code but at the moment it doesn't trigger any errors.


RenGusta:
thanks bae

